PR #11, opened 2019-03-07
Setup Package

kimt33:
I think this is standard setup for python packages:
- setuptools for setup.py
- CONTRUTING.md
- tox.ini and travis.yml

It is assumed (at the moment) that the package will be compatible with version 3.4, 3.5, and 3.6. I believe that `setuptools.find_packages` is backwards compatible to Python 3.3, so there shouldn't be any problems. There are no plans to be Python 2.7 compatible. Following linters were selected:
- flake8 plus pydocstyle and import-order
- pylint
- pep8 naming
- bandit


kimt33:
[Setup repository](https://trello.com/c/soqpa2iB/37-setup-repository)

kimt33:
The pytest linter fails because there are no tests. This shouldn't be an issue.

kimt33:
I am merging this.

PR #16, opened 2019-03-12
Add components of angular momentum

kimt33:
The code was extracted from PR #15 (https://github.com/theochem/gbasis/pull/15),
specifically, commit ed51b6e1d0237096b767a1cdbff085cc84d7fe51 by Xiamin Huang

"Components" of an angular momentum is a (a_x, a_y, a_z) where l = a_x + a_y + a_z.

### DIFF
--- a/gbasis/contractions.py
+++ b/gbasis/contractions.py
@@ -26,7 +26,7 @@ class ContractedCartesianGaussians:
         Angular momentum of the set of contractions.
         .. math::
 
-            \sum_i \vec{a} = a_x + a_y + a_z
+            l = \sum_i \vec{a} = a_x + a_y + a_z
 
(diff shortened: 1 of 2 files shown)


kimt33:
[Add property that returns the `\vec{a}`](https://trello.com/c/9eRX2weX/45-add-property-that-returns-the-veca)

kimt33:
The only comment we can make here is that (1) the list might not be correct data type and that (2) the ordering for the angular momentum components is arbitrary. For (1), it doesn't really matter considering how small this list will be in comparison to the others. For (2), the ordering can be changed to whatever order as long as we don't hard code the ordering. 

To propose a different ordering, raise an issue and/or make a pull request.

PR #17, opened 2019-03-12
Add function to compute the normalization constant for a Cartesian Gaussian primitive

wilhadams:


wilhadams:
[Add function for evaluating normalization constant for cartesian Gaussian](https://trello.com/c/OPZ0q7FQ/8-add-function-for-evaluating-normalization-constant-for-cartesian-gaussian)

wilhadams:
[Add function for evaluating normalization constant for cartesian Gaussian](https://trello.com/c/OPZ0q7FQ/8-add-function-for-evaluating-normalization-constant-for-cartesian-gaussian)

gbasis/contractions.py
```diff
@@ -247,3 +250,36 @@ def coeffs(self, coeffs):
...
+    Returns
+    -------
+    norm : float
+        The normalization constant of the Cartesian Gaussian primitive.
+
+    """
+    return (
+        ((2 * exponent / pi) ** (3 / 4))
+        * ((4 * exponent) ** (np.sum(components) / 2))
+        / (sqrt(np.prod(np.array([factorial2(2 * a - 1) for a in components]))))
```
kimt33: Try to support numpy operations. This would mean we want to make functions that can utilize arrays as well as floats.
In this case, `sqrt` should be replaced with `np.sqrt` and `factorial` applied directly to a numpy array (returning a numpy array) rather than using it within list comprehension (then converting it to numpy array).

tests/test_contractions.py
```diff
@@ -168,3 +168,14 @@ def tests_init():
     assert test._charge == 0
     assert np.allclose(test._coeffs, np.array([1, 2, 3, 4]))
     assert np.allclose(test._exps, np.array([5, 6, 7, 8]))
+
+
+# ToDo: Test norm using actual integrals
+def test_cartesian_gaussian_norm():
+    """Test cartesian_gaussian_norm."""
+    assert np.isclose(
+        [cartesian_gaussian_norm(np.array([0, 0, 0]), 0.25)], [0.2519794355383807303479140]
```
kimt33: Could you cite your sources if you got these numbers from somewhere? If it is a website, a link would be helpful.

wilhadams: These values were determined completely arbitrarily and computed by hand.

gbasis/contractions.py
wilhadams: Fixed in latest commit.

kimt33:
I've added the norm function to the `ContractedCartesianGaussians` class and vectorized it.

kimt33:
Needs more testing, but merging it anyways.

PR #25, opened 2019-03-19
Add low level function for computing the multipole moment integrals

kimt33:
What:
Add _compute_multipole_moment_integrals for computing multipole moment integrals
of different orders for a set of contractions with different angular momentums.

Why:
This function will be used to compute the multipole moments of arbitrary order
which can then be used to compute useful quantities such as overlap and kinetic
energy.

Why Specifics:
1. We do not intend for this function to be the memory bottleneck. The size of
the multidimensional array, `overlaps`, will likely be limited by the
contractions that can be grouped together by their coordinates, exponents, and
coefficients. It doesn't seem likely that neither of the maximum angular
momentum component nor the maximum order of the moment to reach a large enough
number to have memory problems (they will probably be almost always less than
20).
2. We aim to exploit vectorization rather than minimizing the number of
operations. This means that we will likely compute and store many more values
than the minimum necessary. The hope is that the optimization that results from
vectorization will offset the cost of redundant evaluations. Considering that the
optimal path (of recursion) is not trivial to find, and that this path is likely
different depending on different cases (angular momentum vectors and orders of
moments), and that this process will need to be repeated for many different
cases, and that Python is not terribly amenable to repeated for loop structures,
vectorization (via numpy) seems to be a good choice.
3. We pick the following recursion path: first, we recurse over the angular
momentum on the left primitive/contraction until the maximum angular momentum
value (out of all three dimensions) of the contractions on the left side; then,
we recurse over the angular momentum on the right primitive/contraction until
the maximum angular momentum value of the contractions on the right side; and
then we recurse over the order of the moment until the maximum moment
component (out of all three dimensions).
4. We store the integrals for all the values until the maximum angular momentum
and the maximum order of moment. Again, memory is not an issue. Brute force
storage made things easier to code. Also, computation of higher order moments
often requires computation of lower order moments. We will get these integrals
for free.
5. Numpy doesn't play too nicely with recursions. For loops were used instead of
some fancy structure. If necessary, numba should do a nice job speeding things
up.
6. The order of the axis in the intermediate quantity, `integrals`, was
determined in part by trial and error and intuition. First three correspond to
the reverse order of the recursion (moment, angular momentum, angular momentum)
because c-contiguous ordering results in faster lookup (and assignment) for axis
on the left side. The next axis corresponds to dimensions (x, y, z) because
`numpy.prod` seems to be faster over axis in the middle. The last two
axis (primitive, primitive) is better for contraction (for `numpy.tensordot`)

### DIFF
--- a/gbasis/moment_int.py
+++ b/gbasis/moment_int.py
@@ -12,10 +12,12 @@ def _compute_multipole_moment_integrals(
     angmoms_a,
     alphas_a,
     coeffs_a,
+    norm_a,
     coord_b,
     angmoms_b,
     alphas_b,
     coeffs_b,
+    norm_b,
 ):
     """Return the multipole moment integrals of two contraction.
 
@@ -37,6 +39,8 @@ def _compute_multipole_moment_integrals(
         Values of the (square root of the) precisions of the primitives on the left side.
     coeffs_a : np.ndarray(K_a,)
         Contraction coefficients of the primitives on the left side.
+    norm_a : np.ndarray(L_a, 3, K_a)
+        Normalization constants of the components (x, y, z) of the contractions on the left side.
     coord_b : np.ndarray(3,)
         Center of the contraction on the right side.
     angmoms_b : np.ndarray(L_b, 3)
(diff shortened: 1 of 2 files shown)


kimt33:
[Add general structure for integrating multipole moments](https://trello.com/c/D54WnHCk/5-add-general-structure-for-integrating-multipole-moments)

kimt33:
https://trello.com/c/D54WnHCk/5-add-general-structure-for-integrating-multipole-moments

kimt33:
Needs test against reference values.

kimt33:
This function is not tested against references (i.e. other integration modules) but I will merge it anyways so we can make progress on other things.

PR #28, opened 2019-03-21
Vectorize hermite polynomial evaluation in `_eval_deriv_contractions`

kimt33:
`numpy.polynomial.hermite.hermval` was previously used to evaluate the Hermite
series. It had an API that made it difficult to vectorize without computing
much more values than needed (and extracting the desired values as a pain). This
resulted in 3 nested for loops that seemed to be the bottleneck of the code.

`scipy.special.eval_hermite` is used instead to evaluate all of the Hermite
polynomials at the given points for the different orders. Then these values are
linearly combined using the coefficients. This approach was found to be
significantly faster than the previous approach.

Issues
---------
I'm not sure if `scipy.special_eval_hermite` uses some smart
vectorizing/caching to evaluate orders in sequence. It probably doesn't,
considering how arbitrary degrees of the polynomial can be provided. This would
mean that a smarter approach to evaluating Hermite polynomial may speed up the
code, though I don't know if this part is still the bottleneck of the
_eval_deriv_contractions method.

### DIFF
--- a/gbasis/deriv.py
+++ b/gbasis/deriv.py
@@ -1,7 +1,7 @@
 """Derivative of a Gaussian Contraction."""
 from gbasis.contractions import ContractedCartesianGaussians
 import numpy as np
-from scipy.special import comb, perm
+from scipy.special import comb, eval_hermite, perm
 
 
 # TODO: in the case of generalized Cartesian contraction where multiple shells have the same sets of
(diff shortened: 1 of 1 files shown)


kimt33:
Upon @paul's suggestion, I've also tried using `np.polynomial.hermite.hermval(x, np.identity(x.shape[0]))`:
```python
hermite = np.sum(
    coeffs * np.squeeze(np.polynomial.hermite.hermval(alphas ** 0.5 * nonzero_rel_coords,
                                                      np.identity(indices_herm.size)), 1), axis=0
)
```
The `scipy.special.eval_hermite` was faster.

Following code was used to test:
```python
import timeit
    dim = 30
    print(timeit.timeit(
        'eval_hermite(np.arange(dim)[:, None, None, None], x)',
        setup='''
import numpy as np
from scipy.special import eval_hermite
dim = {}
x = np.random.rand(1, dim, dim, dim)
coeffs = np.random.rand(dim, dim)
        '''.format(dim),
        number=10
    ))
    print(timeit.timeit(
        'np.squeeze(np.polynomial.hermite.hermval(x, np.identity(dim)), axis=1)',
        setup='''
import numpy as np
dim = {}
from scipy.special import eval_hermite
x = np.random.rand(1, dim, dim, dim)
coeffs = np.random.rand(dim, dim)
        '''.format(dim),
        number=10

    ))
```

PR #29, opened 2019-03-21
Add function for evaluating contracted spherical Gaussians

wilhadams:


wilhadams:
[Add function for evaluating contracted spherical Gaussian](https://trello.com/c/VAFHzbRu/31-add-function-for-evaluating-contracted-spherical-gaussian)

kimt33:
I made some corrections: mainly that the "contracted Cartesian Gaussian primitives" to "contracted Cartesian Gaussian functions". I thought that the use of primitives was confusing, especially since primitives and contractions are separated. I get that contracted is an adjective that states that the Cartesian Guassian primitives are linearly combined, but the common usage seems more to be "contracted Gaussian type functions".

It really doesn't matter, but I mindlessly changed it and force pushed it. My bad. I will merge it for now, but if you have a problem with this, I can go back and fix it.

PR #33, opened 2019-03-26
Class structure for different types of Gaussian related arrays

kimt33:
The general (naive) approach to computing any Gaussian related arrays, e.g. evaluation, derivatization, density, overlap, electron-nuclear attraction integral, electron-electron repulsion integral, seems to have similar procedures:
1. Compute an element of this array by computing the appropriate integral/evaluation/operation with the appropriate number of primitives. This means that each element of the array can be associated with different numbers of primitives.
2. Build up the entire array for all combinations of primitives
3. Transform the array with respect to primitives to that of contracted Cartesian Gaussian-type functions
4. Transform the array with respect to contracted Cartesian Gaussian-type functions to that of contracted spherical Gaussian-type functions
5. Transform the array with respect to contracted spherical Gaussian-type functions to that of linear combinations of contracted spherical Gaussian-type functions.
6. Apply the appropriate operations to the array.

Instead of this procedure, we modify the first three steps so that we compute an element of the array with respect to the contracted Cartesian Gaussian-type functions that have the same angular momentum at the same center with the same primitive coefficients and exponents. Then, we can vectorize this step to compute elements of multiple contractions at the same time, rather than one at a time. We will call this step 3a. Afterwards, we group all of the arrays associated with each "type" of contraction (shell?). We will call this step 3b. Afterwards, we can continue to step 4.

At the moment, **we will assume that all of the arrays produced by this module will follow this procedure**. This means that all of the arrays with the different number of associated primitives (or contractions) will have a similar structure for steps 3a, 3b, 4 and 5. This pull request contains a tentative class structure that enforces this procedure.

What
--------
1. Add base class `BaseGaussianRelatedArray` 
2. Add base class `BaseOneIndex`
3. Add base class `BaseTwoIndexAsymmetric`
4. Add base class `BaseTwoIndexSymmetric`
5. PyLint changes

Why
------
1. The base class `BaseGaussianRelatedArray` contains `__init__` which is the common initialization procedure for different types of arrays, and abstract methods `construct_array_contraction`, `construct_array_cartesian`, `construct_array_spherical`, and `construct_array_spherical_lincomb` which ensures that its children have the functions necessary for the procedure step 3a, 3b, 4, and 5, respectively.
2. The base class `BaseOneIndex` defines the appropriate transformations an array whose first index is associated with the primitive/contraction. The `construct_array_contraction` still needs to be defined for its children to be instantiated.
3. The base class `BaseTwoIndexAsymmetric` defines the appropriate transformations an array whose first and second indices are associated with two different sets of primitives/contractions. The `construct_array_contraction` still needs to be defined for its children to be instantiated.
4. 3. The base class `BaseTwoIndexAsymmetric` defines the appropriate transformations an array whose first and second indices are associated with the same set of primitives/contractions. The `construct_array_contraction` still needs to be defined for its children to be instantiated.

Issues
---------
Apart from the `BaseTwoIndexSymmetric` which will have one child for the moment integral and another for the electron-nuclear attraction integral, each base class will likely only have one child. This may suggest that all of these structures are unnecessary and confuse users even more. They can, however, be used for different structures, but I'm not sure how likely that will be.

### DIFF
--- a/gbasis/base.py
+++ b/gbasis/base.py
@@ -82,6 +82,16 @@ class BaseGaussianRelatedArray(abc.ABC):
         array_contraction : np.ndarray
             Array associated with the given instance(s) of ContractedCartesianGaussians.
 
+        Notes
+        -----
+        The next level of classes will be divided by number of indices associated with
+        ContractedCartesianGaussians. Then this method's parameters will likely be different from
+        it's children. This means that using this method's API blindly (by simply copying and
+        pasting) will not be suitable for all its children. Here, we allow arbitrary number of
+        `contractions` to indicate that its children may have different number of `contractions` in
+        its parameters. However, in actual implementations, the number of `contractions` will likely
+        be fixed, in which case, an arbitrary number of `contractions` should not be accepted.
+
         """
 
     @abc.abstractmethod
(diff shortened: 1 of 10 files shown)


kimt33:
[Add function for evaluating linear combination of orbitals](https://trello.com/c/64MlosCA/32-add-function-for-evaluating-linear-combination-of-orbitals)

kimt33:
[Make classes for handling different kinds of manipulated Gaussian evaluations/integrations](https://trello.com/c/yk3SpYwg/48-make-classes-for-handling-different-kinds-of-manipulated-gaussian-evaluations-integrations)

PR #34, opened 2019-04-03
Add generalized contractions

kimt33:
Support for generalized contractions (contractions with the same exponents and angular momentum, but different coefficients) have been added.

At the low level (`gbasis.deriv._eval_deriv_contractions`), there was not much of a difference to the code. 

We didn't make the generalized contraction the default for all of the contractions because the segmented contractions (which should be more commonly used) will lose some (minor) performance. To do so, we needed to assume that the abstract method `construct_array_contraction` returns a specific shape for the output. Maybe it'll be better to assume that all contractions are generalized contractions.

### DIFF
--- a/gbasis/base_one.py
+++ b/gbasis/base_one.py
@@ -77,8 +77,12 @@ class BaseOneIndex(BaseGaussianRelatedArray):
 
         Returns
         -------
-        array_contraction : np.ndarray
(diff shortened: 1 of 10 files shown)


kimt33:
[Support generalized contractions](https://trello.com/c/BIeTBMHp/49-support-generalized-contractions)

PR #45, opened 2019-04-26
Add general structure for one-electron integrals

wilhadams:


wilhadams:
[Add general structure for one-electron integrals](https://trello.com/c/6bWGOXp4/6-add-general-structure-for-one-electron-integrals)

kimt33:
So I just made a big push. Not all of these commits are directly related to the point charge integral, they were needed for testing (especially since the reference for testing was the nuclear electron attraction integral). Other nonrelated changes are cosmetic.

There were some changes to the structure, moving parts around and rename variables and other API changes. The critical changes are the bug fixes:
1. normalization was missing (this also makes all of the handwritten tests unusable)
2. the left and right orbitals were not swapped back after swapping
3. missing negative sign
4. some unknown bug in recursion. maybe in the horizontal part at the third level?

I couldn't figure out the bug in the recursion so I rewrote it to be simpler. I think it's somewhere in the horizontal recursion at the third level, but I'm not 100% sure because each component of the code is not explicitly tested. We will need the unit tests for this module (`_one_elec_int`) at one point, but at least we have and pass some tests against references. 

In the future, we will need to make unit tests for `_one_elec_int` and improve the recursion algorithm (current implementation should be slower than the one before).

kimt33:
[Add integral for electron-nuclear attraction](https://trello.com/c/5erbkuZk/24-add-integral-for-electron-nuclear-attraction)

kimt33:
[Add integral for electrostatic potential](https://trello.com/c/xFA4Z97T/23-add-integral-for-electrostatic-potential)

PR #54, opened 2019-05-25
Support basis sets with mix of Cartesian and spherical contractions

kimt33:
Turns out that this is a common enough occurrence to support it. 

I don't think that this API is the ideal one, there's quite a bit of repetition. One case is that the most general case `construct_array_mix` can be used to substitute `construct_array_spherical` and `construct_array_cartesian` with the appropriate keyword arguments. However, I didn't replace these methods in case the `construct_array_mix` will be more difficult for JIT compilers (and other optimization techniques) to optimize the code. If not, we can remove them in the future. If so, I think that the fully cartesian and fully spherical cases are more common occurrences that we can make a separate (repeated) code for each of them.

In this PR, the methods `construct_array_mix` and  `construct_array_lincomb` were added, and `construct_array_spherical` was removed. Some of the docstrings and tests have been updated.


kimt33:
Failed test will be overwritten by the next PR (for esp). The electrostatic_potential.py was accidentally added. It's not even being tested in this commit.

PR #56, opened 2019-05-27
Add kinetic energy integral

kimt33:
1. Split out overlap from _moment_int._compute_multipole_moment
2. Add recursion for integral over orders of a differential operator
3. Add KineticEnergyIntegral class
4. Add helper functions
5. Add tests. Compare against HORTON results
ab1710a


kimt33:
[Add kinetic energy integral using derivative](https://trello.com/c/e1t4tYAb/22-add-kinetic-energy-integral-using-derivative)

PR #57, opened 2019-05-27
Change ordering of the angular momentum components of Cartesian contractions

kimt33:
Before, the angular momentum components of Cartesian contractions were ordered
by giving precedence to z component, then y component, then x component so that
z component has the highest value first, then y and so on. However, this seemed
a little counter intuitive seeing as we normally count from x to z. The ordering
was changed to give precedence to x, y, then z.

What:
1. Change default ordering of the angular components in
ContractedCartesianGaussians
2. Update tests

### DIFF
--- a/gbasis/contractions.py
+++ b/gbasis/contractions.py
@@ -315,8 +315,8 @@ class ContractedCartesianGaussians:
         return np.array(
             [
                 (x, y, self.angmom - x - y)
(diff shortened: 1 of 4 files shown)


kimt33:
[Change default ordering for cartesian contractions](https://trello.com/c/VFbCVxsg/54-change-default-ordering-for-cartesian-contractions)

PR #59, opened 2019-05-31
Add low level  two-electron integrals

kimt33:
1. Add `two_int_brute`. This is a brute force implementation of the recursion in two-electron integrals. No vectorization.
2. Add `_two_elec_int._compute_two_elec_integrals_angmom_zero`. This computes the two electron integrals when all four contractions have angular momentum of zero.
3. Add `_two_elec_int._compute_two_elec_integrals`. This computes the two electron integrals for arbitrary angular momentums (except when all contractions have angular momentum of zero).

So lots of trial and error here. The main problem was that the straight forward vectorized approach resulted in the creation of an array that is way too big for larger angular momentums. Since this is a multidimensional array that is being created, memory usage scales very badly with the angular momentum and number of primitives. I did what I could with the scaling problems with angular momentum (we can reach angular momentum 7 for one contraction), but I haven't really checked the problems associated with a larger number of primitives. I think it shouldn't be too bad since the arrays associated with the angular momentum (in horizontal recursion) scales much worse than the number of primitives, but can't be too sure. For now, we'll address the issue when it becomes a problem.

The case with zero angular momentum contractions is treated separately mainly due to the problem with issues with broadcasting two arrays with zero elements into one another. For the moment and one electron integrals, we could use slicing to treat the zero angular momentum cases, but that doesn't seem possible in the electron transfer recursion. It is my belief that inserting an if statement here may cause trouble for optimization in the future (numba maybe?). I'm not 100% sure if this is true, but we can check later when we optimize the code later.

### DIFF
--- a/gbasis/_two_elec_int.py
+++ b/gbasis/_two_elec_int.py
@@ -379,24 +379,26 @@ def _compute_two_elec_integrals(
     )
     integrals_cont = np.tensordot(integrals_cont * norm_d, coeffs_d, (6, 0))
 
-    # NOTE: Ordering convention for horizontal recursion (d) of integrals
+    # NOTE: Ordering convention for horizontal recursion of first and second indices of d
     # axis 0 : d_x (size: angmom_d + 1)
     # axis 1 : d_y (size: angmom_d + 1)
-    # axis 2 : d_z (size: angmom_d + 1)
-    # axis 3 : c_x (size: m_max_c)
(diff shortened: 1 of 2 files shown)


kimt33:
[Add general structure for two-electron integrals](https://trello.com/c/zeQYB9Cm/7-add-general-structure-for-two-electron-integrals)

PR #60, opened 2019-06-01
Add two electron repulsion

kimt33:
1. Add BaseFourIndexSymmetric class
2. Add ElectronRepulsionIntegral class
3. Make note of bug.

During the recursion, if the exponents are large enough, the intermediate values overflow, resulting in a loss in accuracy. Not sure how to address this, but a test was added.


kimt33:
[Add electron-electron repulsion integral](https://trello.com/c/pXSJ6dtY/25-add-electron-electron-repulsion-integral)

PR #61, opened 2019-06-02
Add general form of the kinetic energy density

kimt33:
No tests.


kimt33:
[Add function for evaluating the general form of kinetic energy density](https://trello.com/c/tXV7L4jH/20-add-function-for-evaluating-the-general-form-of-kinetic-energy-density)

kimt33:

PR #63, opened 2019-06-02
Add integral over angular momentum operator 

kimt33:
No tests.


kimt33:
[Add function for integral over angular momentum](https://trello.com/c/llqTJlci/58-add-function-for-integral-over-angular-momentum)

kimt33:

PR #64, opened 2019-06-02
Add integral over momentum operator

kimt33:
Issues:
No tests


kimt33:
[Add function for integral over momentum](https://trello.com/c/fItUYntN/57-add-function-for-integral-over-momentum)

kimt33:

PR #68, opened 2019-06-04
Reformat docstrings for Napoleon compliance

wilhadams:
Currently there are some docstring section headers that aren't recognized by Napoleon/numpydoc guidelines.

### DIFF
--- a/gbasis/base_one.py
+++ b/gbasis/base_one.py
@@ -131,7 +131,7 @@ class BaseOneIndex(BaseGaussianRelatedArray):
         for contraction in self.contractions:
             array = self.construct_array_contraction(contraction, **kwargs)
             # normalize contractions
-            array *= contraction.norm_cont.reshape(*array.shape[:2], *[1 for i in array.shape[2:]])
+            array *= contraction.norm_cont.reshape(*array.shape[:2], *[1 for _ in array.shape[2:]])
             # ASSUME array always has shape (M, L, ...)
             matrices.append(np.concatenate(array, axis=0))
         return np.concatenate(matrices, axis=0)
@@ -165,7 +165,7 @@ class BaseOneIndex(BaseGaussianRelatedArray):
             matrix_contraction = self.construct_array_contraction(cont, **kwargs)
             # normalize contractions
             matrix_contraction *= cont.norm_cont.reshape(
-                *matrix_contraction.shape[:2], *[1 for i in matrix_contraction.shape[2:]]
+                *matrix_contraction.shape[:2], *[1 for _ in matrix_contraction.shape[2:]]
             )
             # transform
             # ASSUME array always has shape (M, L, ...)
@@ -228,7 +228,7 @@ class BaseOneIndex(BaseGaussianRelatedArray):
             matrix_contraction = self.construct_array_contraction(cont, **kwargs)
             # normalize contractions
             matrix_contraction *= cont.norm_cont.reshape(
-                *matrix_contraction.shape[:2], *[1 for i in matrix_contraction.shape[2:]]
+                *matrix_contraction.shape[:2], *[1 for _ in matrix_contraction.shape[2:]]
             )
             # transform
             # ASSUME array always has shape (M, L, ...)
--- a/gbasis/base_one.py
+++ b/gbasis/base_one.py
@@ -248,7 +248,6 @@ class BaseOneIndex(BaseGaussianRelatedArray):
 
             \sum_{j} T_{i j} M_{jklm...} = M^{trans}_{iklm...}
 
-
         Parameters
         ----------
         transform : np.ndarray(K_orbs, K_cont)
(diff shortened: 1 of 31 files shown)


kimt33:
The changes also include:
1. Fix spelling/grammar in docstring
2. Make docstrings consistent for repeating patterns
3. Add back quotation for references to python objects
4. Use dimension i instead of axis i

gbasis/base_one.py
```diff
@@ -17,34 +17,33 @@ class BaseOneIndex(BaseGaussianRelatedArray):
...
     -------
     __init__(self, contractions)
         Initialize.
     construct_array_contraction(self, contraction) : np.ndarray(M, L_cart, ...)
         Return the array associated with a `GeneralizedContractionShell` instance.
-        `M` is the number of segmented contractions with the same exponents (and angular momentum).
-        `L_cart` is the number of Cartesian contractions for the given angular momentum.
+        :math:`M` is the number of segmented contractions with the same exponents (and angular
+        momentum).
+        :math:`L_cart` is the number of Cartesian contractions for the given angular momentum.
```
kimt33: So the whole point of `:math:` is to let sphinx  (or whatever renderer) know to render the content as a latex equation rather than to let the user know that it is a math variable. In this case, `L_cart` as a latex equation will look like the following 
https://latex.codecogs.com/gif.latex?L_cart
which will need to be fixed when we start using sphinx.

Now I realized this change came about because the form of the docstring was being standardized, and some of the variables are treated as math equations and some treated as "code". I like the use of math equations as much as possible since it makes as consistent (the math equations are rendered differently from backquoted objects) with the equations that are present elsewhere, but this case doesn't really fit, unless you change the whole thing to ":math:`L_{\mathrm{cart}}`" which is a bit of mouthful. Also these terms will be rendered differently from `np.ndarray(M, L_cart, ...)`.

I think it'd be better if the `:math:` part is dropped here.

Also, I don't remember sure how much experience you have with Latex (I remember talking to you about it but I forgot), but if you aren't super comfortable with it, we can treat it as a normal code object, as fix it later when we start generating the html's.

kimt33 (changes requested):
I think apart from the one change, everything looks good.

The change I requested, however, is a bit of pain to fix. So I don't know how you'd like to proceed. I think you can remove all references to `:math:`, it would be hard to notice the unnecessary change. To make it easier to notice, if you do do this, I think it'd be best if you make a new commit instead of ammending.

FarnazH:
@kimt33 I will be adding small commits so that it is easy to undo. Please let me know if you think sth should be reversed.

gbasis/base_one.py
wilhadams: I had meant to only apply `:math:` to the math usages of variables, not the "code" usages, but I guess I was on autopilot and ended up losing sight of which ones I meant to change. 

I'm comfortable with Latex, but this is my first time using Sphinx, so for some reason that's been tripping me up, even though it's just being treated as a Latex equation.

gbasis/evals/electrostatic_potential.py
```diff
@@ -13,23 +13,25 @@ def electrostatic_potential(
...
-        be expressed with respect to the transformed basis set.
+        If the basis is transformed using `transform` keyword, then the density matrix is assumed
+        to be expressed with respect to the transformed basis set.
     points : np.ndarray(N, 3)
-        Coordinates of the points in space (in atomic units) where the basis functions are
-        evaluated.
-        Rows correspond to the points and columns correspond to the x, y, and z components.
+        Cartesian coordinates of the points in space (in atomic units) where the basis functions
+        are evaluated.
+        Rows correspond to the points and columns correspond to the :math:`x, y, \text{and} z`
```
kimt33, line 29: FYI can be replaced with :math:`x, y,` and :math:`z` also. Or :math:`x`, :math:`y`, and :math:`z`

gbasis/evals/electrostatic_potential.py
```diff
@@ -53,33 +55,35 @@ def electrostatic_potential(
...
+        Electrostatic potential evaluated at the given points.
 
     Raises
     ------
     TypeError
-        If `one_density_matrix` is not a two-dimensional numpy array.
-        If `nuclear_coords` is not a two-dimensional numpy array with 3 columns.
-        If `nuclear_charges` is not a one-dimensional numpy array.
-        If `threshold_dist` is not a int/float.
+        If `one_density_matrix` is not a two-dimensional `numpy` array.
```
kimt33, line 63: I don't think numpy here needs to be backquoted.

gbasis/base_two_symm.py
```diff
@@ -30,14 +28,14 @@ class BaseTwoIndexSymmetric(BaseGaussianRelatedArray):
     construct_array_contraction(self, contraction, **kwargs) :
     np.ndarray(M_1, L_cart_1, M_2, L_cart_2, ...)
         Return the array associated with a `GeneralizedContractionShell` instance.
-        `M_1` is the number of segmented contractions with the same exponents (and angular momentum)
+        `M_1` is the number of segmented contractions with the same exponents (and angular
+        momentum) associated with the first index.
```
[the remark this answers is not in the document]
kimt33: Nevermind

kimt33:
Thank you all for your help!

PR #69, opened 2019-06-04
Add interface to iodata module

kimt33:
Issues:
1. I cannot understand the convention used for the spherical contraction/primitive
ordering.
2. Didn't want dependency on iodata so the input is checked by class and
attribute names
3. The API of iodata used has not been updated on pypi nor on github (not
tagged).


kimt33:
[Add wrapper to iodata](https://trello.com/c/Z5mAU4Ek/53-add-wrapper-to-iodata)

tovrstra:
@kimt33 I haven't been able to update the docs on RTD. It is under the account of @matt-chan and it seems to be non-trivial to get this to work in short term. For Pypi, we would have to make a release, which can be done as soon as we feel IOData is mature enough for it.

kimt33:
@tovrstra I understand. That's sort of what I figured. Sorry to derail from the topic but could you have a look at https://github.com/theochem/iodata/issues/121? I'm having trouble figuring out the ordering of the spherical contractions.

kimt33:
I implemented what I think is the correct ordering. See https://github.com/theochem/iodata/issues/121 for explanation. Thank you @tovrstra

PR #70, opened 2019-06-04
Add parser for gbs format

kimt33:
Issues:
Gaussian94 basis format does not explicitly state which contrations are
generalized (unlike the nwchem format). This means that we must infer which
contractions belong to the same generalized shell. We can treat all of the
contractions to be segmented then collect them together afterwards, but we will
end up losing information on the ordering. I'm guessing that if someone decides
to separate out the contractions of a generalized shell, they must've done it
for a reason and that reshuffling them together will cause confusion. Therefore,
the order is preserved and contractions are treated to be segmented.

The generalized contraction shell as implemented in this module does not support
more than one angular momentum


kimt33:
[Add parser for gbs format](https://trello.com/c/lUzYDraz/67-add-parser-for-gbs-format)

PR #71, opened 2019-06-04
 Add interface to pyscf

kimt33:
Issue:
Couple hacky things:
1. I didn't want to have pyscf dependency, so the input is checked by the class
and attribute names
2. Tox doesn't install pyscf, which means that this part won't be tested by
travis (but it is tested with 100% coverage)


kimt33:
[Interface to Pyscf](https://trello.com/c/nlleLSCC/69-interface-to-pyscf)

PR #84, opened 2019-06-06
Restructure code

kimt33:
From the documentation, it's clear that there are two types of ways that this module will be used:
1. Evaluate some property related to the Gaussians
2. Integrate Gaussians together with some operator

From an organization perspective, structuring the code to reflect these use cases seems to make sense. However, from a user perspective, it seems to add a little bit of baggage. For example, to get nuclear-electron attraction integral we go from
```python
from gbasis.nuclear_electron_attraction import nuclear_electron_attraction_integral
```
to
```python
from gbasis.integrals.nuclear_electron_attraction import nuclear_electron_attraction_integral
```
Maybe it's not that much since "integrals" does make up most of the text.
One case, however is a little bit iffy though
```python
from gbasis.evals.eval import evaluate_basis
```
But all these seem to have to do with the names of the modules rather than their positions, which can be a separate issue/PR altogether.

## Type of Changes

|   | Type |


kimt33:
@PaulWAyers @wilhadams any thoughts? objections?

kimt33:
[Add directory structure](https://trello.com/c/Ey8LaD6n/76-add-directory-structure)

wilhadams:
I think this is a nice way to separate things. It may get a little tedious to import things, and we can definitely take a look at module names and make improvements where needed, but I think the extra organization will be nice. 

Either way, IDEs like PyCharm will autocomplete the module names, so for those users the added burden is negligible.

Issue #78, opened 2019-06-06
Bad iodata support


kimt33:
The module iodata is supported, technically, but not very well. `iodata` for which the interface was created is not available on neither pypi nor conda. It was designed for the latest iodata in the master branch of the repository https://github.com/theochem/iodata. There are some major API breaking changes from the previous version which makes the current wrapper not compatible with the ones in pypi or conda. 

The setup.py references that the iodata must be greater than 0.1.7 (current version in pypi and conda), but due to reasons the latest iodata has not been pushed or versioned yet. 

The fix, at the moment, is to clone the repository and install it manually
```bash
https://github.com/theochem/iodata
cd iodata
pip install --user -e .
```
and the `pip install gbasis[iodata]` is broken until iodata is available (This command wouldn't work anyways because the current version of gbasis isn't up on pypi or conda, but that'll happen soon).

See PR https://github.com/theochem/gbasis/pull/69 for some discussion

tovrstra:
One can always directly install with pip from a github repo as follows. As long as we have no IOData release, this could be helpful:

```bash
pip install git+https://github.com/theochem/iodata.git@master
```

FarnazH:
What should be done to resolve this error message? https://github.com/theochem/gbasis/blob/master/gbasis/wrappers.py#L108

I was loading a `WFN` which contains only Cartesian primitives, and got this error: 
`ValueError: Given convention does not support spherical contractions for the angular momentum 2`

tovrstra:
@FarnazH @kimt33 I'd like to pick this up again. There are still a few cases for which the wrapper does not work yet. 

I'm doing some prototyping for denspart, just to see which ideas could work and where we might run into problems. I wrote a fairly naive input preparation code (to set up a density on an integration grid) and check for a bunch of test files if the density integrates to the right number of electrons, see https://github.com/theochem/denspart/blob/prototype0/denspart/adapters/test/test_horton3.py

This is the output: 

```
denspart/adapters/test/test_horton3.py::test_integrate_rho[2h-azirine-cc.fchk] PASSED [  1%]
denspart/adapters/test/test_horton3.py::test_integrate_rho[2h-azirine-ci.fchk] PASSED [  2%]
denspart/adapters/test/test_horton3.py::test_integrate_rho[2h-azirine-mp2.fchk] PASSED [  3%]
denspart/adapters/test/test_horton3.py::test_integrate_rho[2h-azirine-mp3.fchk] PASSED [  4%]
[...]

PR #102, opened 2019-08-26
Update iodata convention for pure functions

kimt33:
## Steps

- [x] Rebase onto master

## Description

The angular momentum symbol was removed from the string that is used to describe
a pure function in iodata. This commit ensures that the gbasis module is
compatible with the commit https://github.com/theochem/iodata/commit/4ceb9adbf8dd3438e2e59afd1bed947c5c9abd16 of
iodata (master as of August 26, 2019).

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Issue #78


tovrstra:
Looks good to me.

PR #105, opened 2020-01-29
Fix some minor code formatting issues

tovrstra:
## Steps

- [x] Write a good description of what the PR does.
- [x] Add tests for each unit of code added (e.g. function, class) -> **not needed, just reformatting**
- [x] Update documentation -> **not needed, just reformatting**
- [x] Squash commits that can be grouped together
- [x] Rebase onto master

## Description

There are a few small formatting issues that make the tests on Travis-CI fail for the current master branch. This PR fixes them. I guess black and pylint have changed their default behavior recently.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :toilet: Cleaning  |

## Related Issue

I ran into these when working on #78 but these changes are unrelated.

PR #106, opened 2020-01-29
Make the IOData wrapper work for most cases.

tovrstra:
## Steps

- [x] Update documentation
- [x] Rebase onto master

## Description

- Add a return value to from_iodata, the make sure the caller gets a complete description of the basis set.
- Removed hard-coded conventions for spherical functions, no longer needed.
- Set all normalization constants for the contractions to 1.0. IOData does not impose this convention.
- Just use some convention for Cartesian functions when this info is not provided by IOData. Which is used does not matter.
- Remove redundant arguments from _replace call.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

This fixes a part of #78.

### DIFF
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -16,6 +16,11 @@ def from_iodata(mol):
     basis : tuple of gbasis.contraciton.GeneralizedContractionShell
         Basis set object used within the `gbasis` module.
         `GeneralizedContractionShell` corresponds to the `Shell` object within `iodata.basis`.
+    coord_types : list of str
+        A list of strings, each on being either ``"cartesian"`` or
+        ``"spherical"``, which needs to be passed into integral and eval
+        functions. This part of the basis set is not stored in the ``basis``
+        return value. Users need to keep track of this in a separate variable.
 
     Raises
     ------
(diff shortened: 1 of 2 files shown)


tovrstra:
The tests fail only because of previously existing formatting issues, which are fixed in #105.

tovrstra:
I'll rerun the tests after merging the fixes for black. They should pass now, if not I'll fix.

evohringer:
Toon, we are experiencing some problems with basis sets containing f functions in molden file format (more specifically the def2-TZVP basis set from ORCA).
The error is 
```
File "charges_param/gbasis/wrappers.py", line 114, in angmom_components_sph
    "Only the real solid harmonics as defined in Helgaker Section 6.4.2. is "
NotImplementedError: Only the real solid harmonics as defined in Helgaker Section 6.4.2. is supported.
```

 Do you think these changes will fix it?

wilhadams:
@evohringer these changes won't fix the issue, it has to do with the way `gbasis` deals with ordering the pure (spherical) functions, which doesn't yet support ORCA-style conventions. This will be fixed soon.

wilhadams (approved):

PR #107, opened 2020-05-16
Fixed bug in parsers.py which failed to load certain families of basis sets

BradenDKelly:
## Steps

- [x] Update documentation
- [x] Rebase onto master

## Description
When parsing certain families of basis sets, parsers_gbs had a bug which caused parsing to fail. The source of the problem was np.allclose() which compares two arrays element by element. In reality, without the bug, all basis sets would have failed to load, so the bug may actually be in np.allclose(). np.allclose() was used to see if the previous shells exponents were the same as the current shells exponents. If yes, the coefficients were np.hstack together, else, they current shell was appended. For some reason np.allclose() did not raise a flag when previous shell had n>1 exponents and current had n=1 i.e., was a primitive. It should have. Interestingly, for many families it is always the case that a shell with many primitives in the contraction, is followed by a single primitive in the next shell. However, for other families, particularly 6-31, where the first s-orbital has 6 primitives, and the next s-orbital has 3, np.allclose() raised an error (6,) vs (3,) incompatible. I am sure there are other families too.

## Type of Changes
Added simple check prior to using np.allclose() to ensure both arrays were same length. If not, a variable `hstack` is made false and the coefficients are not np.hstack together. If they are the same length, then np.allclose() is used to see if they are identical. if so, the variable `hstack` is made True, else False.

i.e.,

                # dummy fix for np.allclose
                if output[atom] and len(output[atom][-1][1]) == len(exps):
                    hstack = np.allclose(output[atom][-1][1], exps)
                else:
                    hstack = False

|   | Type |
| ✓  | :bug: Bug fix  | True


gbasis/parsers.py
```diff
@@ -99,15 +99,15 @@ def parse_gbs(gbs_basis_file):
...
     data = re.split(r"\n\s*(\w[\w]?)\s+\w+\s*\n", gbs_basis)
     dict_angmom = {"s": 0, "p": 1, "d": 2, "f": 3, "g": 4, "h": 5, "i": 6, "k": 7}
     # remove first part
     if "\n" in data[0]:  # pragma: no branch
         data = data[1:]
-
-    atoms = data[::2]
-    basis = data[1::2]
+    atoms = data[::2]  # stride of 2 get the ['H','C', etc] from e.g., above
+    basis = data[1::2]  # starting with 'stuff' take strides of 2 to skip elements
```
wilhadams: Could you merge these into a single comment above the line? We try to avoid inline comments

gbasis/parsers.py
```diff
@@ -146,18 +147,26 @@ def parse_gbs(gbs_basis_file):
             # if len(angmom_seg) == 1:
             #     coeffs_seg = coeffs_seg[:, None]
             for i, angmom in enumerate(angmom_seg):
+
+                # dummy fix for np.allclose
```
wilhadams: Please make this more descriptive ("check if previous exponents are the same as these exponents")

gbasis/parsers.py
```diff
@@ -146,18 +147,26 @@ def parse_gbs(gbs_basis_file):
...
+                    hstack = False
+
                 if (
-                    output[atom]
-                    and output[atom][-1][0] == angmom
-                    and np.allclose(output[atom][-1][1], exps)
+                        output[atom]
+                        and output[atom][-1][0] == angmom
+                        and hstack
+                        # and np.allclose(output[atom][-1][1], exps)
```
wilhadams: Drop this line

gbasis/parsers.py
```diff
@@ -146,18 +147,26 @@ def parse_gbs(gbs_basis_file):
...
+                else:
+                    hstack = False
+
                 if (
-                    output[atom]
-                    and output[atom][-1][0] == angmom
-                    and np.allclose(output[atom][-1][1], exps)
+                        output[atom]
+                        and output[atom][-1][0] == angmom
+                        and hstack
```
wilhadams: Check `black` output, these are too indented

gbasis/parsers.py
```diff
@@ -146,18 +147,26 @@ def parse_gbs(gbs_basis_file):
...
             #     coeffs_seg = coeffs_seg[:, None]
             for i, angmom in enumerate(angmom_seg):
+
+                # dummy fix for np.allclose
+                if output[atom] and len(output[atom][-1][1]) == len(exps):
+                    hstack = np.allclose(output[atom][-1][1], exps)
+                else:
+                    hstack = False
+
                 if (
```
wilhadams: Maybe comment "check if generalized contraction can be built"

gbasis/parsers.py
```diff
@@ -146,18 +147,26 @@ def parse_gbs(gbs_basis_file):
...
+                        output[atom]
+                        and output[atom][-1][0] == angmom
+                        and hstack
+                        # and np.allclose(output[atom][-1][1], exps)
                 ):
                     output[atom][-1] = (
                         angmom,
                         exps,
-                        np.hstack([output[atom][-1][2], coeffs_seg[:, i : i + 1]]),
+                        np.hstack([output[atom][-1][2], coeffs_seg[:, i:i + 1]]),
```
wilhadams: Please revert, check `black` output

gbasis/parsers.py
```diff
@@ -146,18 +147,26 @@ def parse_gbs(gbs_basis_file):
...
                     )
                 else:
-                    output[atom].append((angmom, exps, coeffs_seg[:, i : i + 1]))
+                    output[atom].append((angmom, exps, coeffs_seg[:, i:i + 1]))
```
wilhadams: Please revert, check `black` output

wilhadams (changes requested):
Good catch on this bug! Please check the "linters" job on Travis, and consider adding the 6-31 basis file to the tests/ dir so we can test to ensure this keeps working.

wilhadams:
Superseded by #109

PR #108, opened 2020-05-26
Adopt cosine- and sine-type notation for spherical ordering

wilhadams:
## Steps

- [x] Update documentation
- [x] Rebase onto master

## Description
Currently in `gbasis` we use the magnetic quantum number to order the pure (spherical)
functions for conversions from cartesian to spherical. This causes problems for cases
where other conventions are used, e.g. ORCA Molden files, which use the negative of
some pure functions. We can prevent this issue and allow for any convention allowed
in `iodata` by adopting the same conventions.

Thus, we'll instead represent the ordering of pure functions using cosine- and sine-like
functions, so the previous notation (-2, -1, 0, 1, 2) corresponds to (s2, s1, c0, c1, c2)
which is consistent with its usage in `iodata`. Further, we now can support ORCA, since
for f-type functions (as an example), ORCA uses the ordering
(c0, c1, s1, c2, s2, -c3, -s3), which requires taking the negative of the real regular
solid harmonic, calculated in `spherical.py`.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

## Related Issue
This will improve `iodata` support (#78) and addresses issues raised in https://github.com/theochem/iodata/issues/130 using the documentation from https://github.com/theochem/iodata/pull/134 

This will also fix the error noted here https://github.com/theochem/gbasis/pull/106#issuecomment-619630911 by allowing for ORCA-style conventions


docs/notes.tex
```diff
@@ -564,6 +564,60 @@ \subsection{Ordering of basis functions}
 \verb|GeneralizedContractionShell|
 and overwrite these properties with the desired ordering.
 
+\subsection{Conventions for ordering Cartesian and spherical components}
+\label{sec:angmom_components_ordering}
+
+\paragraph{Cartesian components}
+A shell of generalized contractions of angular momentum $\ell$ is composed of $(\ell + 1)(\ell + 2)$ Cartesian components of the form $x^{a_x}y^{a_y}z^{a_z}$ such that $a_x + a_y + a_z = \ell$. In \verb|gbasis|, the ordering of Cartesian components is stored as an $\ell \times 3$ \verb|numpy| array, where each entry $(i, j)$ corresponds to the exponent $a_j$ for the $i$\textsuperscript{th} component. In other programs, like \verb|HORTON2|, a different notation is used, where the components are written as strings of X, Y, and Z, where the number of times each letter appears corresponds to the exponent $a$ for that letter. These notations are easily convertible, for example, the \verb|gbasis| notation for the component \verb|XXXZZ| is $(3, 0, 2)$
```
kimt33: I'm not sure what the convention is on wrapping latex, but it might not be a bad idea to separate these sentences into multiple lines. In latex, only double newlines start a new paragraph,

kimt33 (dismissed):
I'll be honest. I didn't look too thoroughly through this PR, but it looks pretty good. 
Not much to say, except for some superficial things.

Also, I think the commits were intended to be cleaned up / rebased but haven't yet been (force) pushed. Or maybe github handles the cleanup for you.

docs/notes.tex
wilhadams: I was writing this in Overleaf and it wraps lines automatically, so I didn't notice. Would definitely be better to break up the long lines.

wilhadams:
> Also, I think the commits were intended to be cleaned up / rebased but haven't yet been (force) pushed. Or maybe github handles the cleanup for you.

Yup, I forgot to do my final rebase, good catch

PR #109, opened 2020-06-02
Fixed bug in parsers.py which failed to load certain families of basis sets

BradenDKelly:
## Steps

- [x] Update documentation
- [x] Rebase onto master

## Description
When parsing certain families of basis sets, parsers_gbs had a bug which caused parsing to fail. The source of the problem was np.allclose() which compares two arrays element by element. In reality, without the bug, all basis sets would have failed to load, so the bug may actually be in np.allclose(). np.allclose() was used to see if the previous shells exponents were the same as the current shells exponents. If yes, the coefficients were np.hstack together, else, they current shell was appended. For some reason np.allclose() did not raise a flag when previous shell had n>1 exponents and current had n=1 i.e., was a primitive. It should have. Interestingly, for many families it is always the case that a shell with many primitives in the contraction, is followed by a single primitive in the next shell. However, for other families, particularly 6-31, where the first s-orbital has 6 primitives, and the next s-orbital has 3, np.allclose() raised an error (6,) vs (3,) incompatible. I am sure there are other families too.

## Type of Changes
Added simple check prior to using np.allclose() to ensure both arrays were same length. If not, a variable `hstack` is made false and the coefficients are not np.hstack together. If they are the same length, then np.allclose() is used to see if they are identical. if so, the variable `hstack` is made True, else False.

i.e.,

                # dummy fix for np.allclose
                if output[atom] and len(output[atom][-1][1]) == len(exps):
                    hstack = np.allclose(output[atom][-1][1], exps)
                else:
                    hstack = False

|   | Type |
| ✓  | :bug: Bug fix  | True


wilhadams:
Supersedes #107

wilhadams (approved):
Looks good, thanks @BradenDKelly

PR #110, opened 2020-06-21
Added overlap screening to gbasis

BradenDKelly:
## Steps

- [ ] Update documentation
- [x ] Squash commits that can be grouped together
- [x ] Rebase onto master

## Description
Overlap screening was added in order to speed up the calculation of overlap integrals by neglecting overlaps that do not contribute significantly.

The screens consist of True/False masks. Each contraction has a boolean array denoting if it does or does not interact with another contractions. 

Inside contractions.py the `class GeneralizedContractionShell` had a method (`def create_overlap_mask(self, basis, eps=1e-20):`) added, which does the grunt working of creating the boolean mask for that contraction. Further, the class `GeneralizedContractionShell` also stores the contraction index, since we need to know if contraction `X` interacts with contraction `Y`.

Inside parsers.py the function `make_contractions(basis_dict, atoms, coords, overlap=False):` was modified to include a call to create overlap masks, if the user wants screening to be done.

```
    if overlap:
        for contractions in enumerate(basis):
            contractions.create_overlap_mask(basis)
```
Inside overlap.py the calls to `_compute_multipole_moment_integrals` were modified so that they are screened. If `contraction X` should not interact with `contraction Y` then a `numpy` matrix of zeros is `returned` rather than performing the overlap calculations.

Tests were implemented in `test_overlap` and `test_parsers`. The tests are for scenarios where overlaps should be done, should not be done, and also, there is a test to make sure that the values with and without screens are actually different at a small enough tolerance. Both Gaussian and nwchem have tests for pairs and triplets of `H` and `C` atom, and for both STO-6g and 6-31G basis sets.

## Type of Changes

|   | Type |
| x  | :bug: Bug fix  |
| ✓  | :scroll: Docs |


tests/data_631g.gbs
```diff
@@ -951,4 +951,4 @@ D   3   1.00
       0.6251213000D+01       0.7120146000D+00
 D   1   1.00
       0.1979236000D+01       1.0000000
-****
+****
```
kimt33, line 954: It seems like this commit was introduced because of the editor stripping the newlines at the end of file.

kimt33 (approved):
Sorry to look at this so late. I haven't been active on the repo for more than a year but I got requested to review this recently so I'm just pushing it through.

My main concerns are that coverage seems to be failing for some reason (for a bunch of the PR's so that might be a repo issue) and that some of the attributes are defined outside the __init__.

I don't think that either should stop the PR from merging, mostly to add critical functionality to the code even if it has to be cleaned a bit in the future. And the tweaks seem easy enough.

I would be a bit more concerned about the coverage failing, but the code seems to test the main functionality of the code so we know it works at least. The added changes are relatively isolated from the other parts of the code so even if bugs are introduced, their effects should be isolated also.

I will approve it, but I would like to leave it up to the current package manager to make the final call. If no one vetoes, I would merge it mostly to add functionality

tests/data_631g.gbs
BradenDKelly: Memory does not serve me well on this one. I don't think we want to merge #110 though, I think that #111 is the better one, it may be worth keeping #110 floating around but the version introduced by #111 is a lot more memory friendly and I believe just simpler in general.

leila-pujal:
PR #111 was merged

PR #111, opened 2020-06-28
Overlap no mask version

BradenDKelly:
## Steps

- [ ] Update documentation
- [x] Rebase onto master

## Description
This is a replacement PR for #110 which also does overlap screening, but is bulkier and uses stored masks which take up more memory.

Screening for overlap integrals was implemented in a simple and concise manner meant to provide a cleaner, significantly less memory intensive overlap screening. For 6-31G* the screening appears to be ~15-20% faster than unscreened for a "polymer" chain of 100 Kr atoms.

The screen is implemented in `overlap.py `  (line 93). Before calculating the moment between two contractions, a function `def calculate_these_overlaps(contractions_one, contractions_two):` is called (line 160) which checks first if screening should be done, and if so, calculates the distance, according to a provided tolerance, at which contractions stop interacting with each other. It returns a true or false value. If overlaps are to be done, they are done - the same as the unscreened code. If they are not to be done, a zero matrix is created that is the same size as would have been generated in the overlap call.

In order to accomplish this some simple book-keeping features are added. The `class GeneralizedContractionShell:`  in `contractions.py` had a tolerance and flag variable added. Both by default are given values, but, the user can provide their own tolerance if they so choose.

The user interacts with `GeneralizedContractionShell` in `parsers.py`, so a slight modification was made here to the function `make_contractions(basis_dict, atoms, coords, tol=1e-20, overlap=False):`, where tol and overlap are added features.

Finally, two unit tests were made in test_overlap.py. They test first that the screened overlaps match the unscreened overlaps to a numpy tolerance of 1e-8 provided the overlap tolerance was 1e-20. Secondly, provided an overlap tolerance of 1e-4, the test is to ensure that the unscreened and screened overlaps do not match within a numpy tolerance of 1e-8.

Actually finally... screening is by default turned off. This is in order to be backwards compatible with the plethora of unit tests already present. This can be changed to be default always being on, provided the unit tests are all updated.
## Type of Changes

|   | Type |
| x | :bug: Bug fix  |
| x  | :scroll: Docs |


kimt33 (dismissed):
I should have looked at this PR before #110. But yeah sorry it's late nonetheless.

My comment here is similar to that of #110. The < 100% coverage thing really should depend on whoever is currently managing the repo. If they're okay with it, it's fine.

One other side comment: the name `compute_these_overlaps` threw me off a little bit for some reason. It makes sense but it seemed like it would return an overlap rather than a condition for computing the overlaps. Not an issue, but yeah.

leila-pujal:
@FarnazH, I took a look again at this PR and everything looks good to me. The tests passe and I looked at the code and it matches Paul's notes. I haven't merged because David raised a concern with the name of one of the functions `compute_these_overlaps`. This returns a boolean and it may look like this function actually compute the overlaps. You are better than me in coming up with new names so maybe you have a good idea of what we should rename the function.

BradenDKelly:
Hey, I recognize this! but doubt I can be of assistance. I think Farnaz is the best person on this particular day! You folks sure are on a coding roll the last week!

PaulWAyers:
We decided to use `is_overlap_included` in the hackathon meeting today.

leila-pujal:
@FarnazH, I pushed a commit changing the name of the function. I could merge this but I will have to deal with the conflicts. I think you have been dealing with these more than me, feel free to merge this if you think it will be easier for you, but I know you have a lot on your plate so feel free to let me know if you want me to take care of the conflicts.

PaulWAyers (approved):
Looks good to me. Thanks @BradenDKelly

PR #118, opened 2021-04-18
Speed up gradient and hessian computation Closes #99

### LINKED ISSUE #99: [REQUEST] Speedup gradient & hessian computation
**Is your feature request related to a problem? Please describe.**

When incorporating `gbaiss` in `chemtools`, I noticed that gradient and hessian computations are much slower than anticipated when compared to `horton2` which is currently used in `chemtools`. In `gbasis`, evaluating density seems to be pretty fast and I was hoping that gradient and hessian evaluations take 3 and 6 times as much. So, I am wondering whether there is room for improving the implementations.

**Is there a specific integral/formula that you would like implemented?**

@PaulWAyers has included a direct implementation of gradient and hessian in his notes.

**Is there a change to the code or algorithm you would like to see?**

If possible, it would be good to have a faster implementation of gradient and hessian computation, by either hard-coding these derivative and/or taking advantage of Hessian symmetries or using `cython`.

**Additional context**

Here are the code snippet and files to time the computations for H2O using cc-pVTZ basis on a cubic grid with 105,840 points (please unzip the attached `files.tar.gz` to get the `h2o.fchk` and `h2o.npz`). On my system, using `time` module, computing density, gradient, and hessian take ~0.8, ~10, ~55 seconds, respectively.  

```python
#!/usr/bin/env python

import time
import numpy as np

from iodata import load_one

from gbasis.wrappers import from_iodata
from gbasis.evals.density import evaluate_density
from gbasis.evals.density import evaluate_density_gradient, evaluate_density_hessian

# load points & density matrix
data = np.load("h2o.npz", allow_pickle=True)
points, dm = data["points"], data["dm"]
print("points shape = ", points.shape)

# build basis & compute
mol = load_one("h2o.fchk")
basis = from_iodata(mol)

start = time.time()
# dens = evaluate_density(dm, basis, points)
# grad = evaluate_density_gradient(dm, basis, points)
hess = evaluate_density_hessian(dm, basis, points)
end = time.time()
print("computation time = ", end - start)
```

[files.tar.gz](https://github.com/theochem/gbasis/files/3355471/files.tar.gz)

leila-pujal:
## Steps

- [x] Update documentation
- [ ] Rebase onto master

## Description
New code to speed up hessian and laplacian calculations.  Two main modifications have been made:
-New gradient, laplacian and hessian calling functions: Old functions computed gradient, laplacian and hessian matrices in a general form that calculated several derivative orders more than once and in some cases symmetry was not used. New functions are implemented to take advantage of this and speed up the code.
-New functions with explicitly coded first and second derivatives: 1st and 2nd derivative equations are coded explicitly in  `_deriv.py` as the main goal was to speed up laplacian and hessian calculations, where the highest order for derivatives is 2. In `_deriv.py` 3 new functions were added, `_eval_first_second_order_deriv_contractions()` with the main code to deal with the calculation and gbasis organization, and `_first_derivative()` and `_second_derivative()` where explicit formulas for  1st and 2nd  from gbasis notes are coded. 

New argument `deriv_type`: A new argument `deriv_type` was introduced to be able to choose between `_eval_deriv_contractions()`- general formula for derivation of contractions, and `_eval_first_second_order_deriv_contractions()` - direct formula for 1st and 2nd derivatives. 

New tests for `_eval_first_second_order_deriv_contractions()`-direct formula for 1st and 2nd derivatives in `test_density_direct.py`. It is the same test module as `test_density.py` but with deriv_type argument set equal to `direct`. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/gbasis/evals/density.py
+++ b/gbasis/evals/density.py
@@ -255,16 +255,28 @@ def evaluate_deriv_density(
                 num_occurence = comb(total_l_x, l_x) * comb(total_l_y, l_y) * comb(total_l_z, l_z)
                 orders_one = np.array([l_x, l_y, l_z])
                 orders_two = orders - orders_one
-                density = evaluate_deriv_reduced_density_matrix(
-                    orders_one,
-                    orders_two,
-                    one_density_matrix,
-                    basis,
-                    points,
-                    transform=transform,
-                    coord_type=coord_type,
-                    deriv_type=deriv_type,
-                )
+                if any(orders_one > 2) or any(orders_two > 2):
+                    density = evaluate_deriv_reduced_density_matrix(
+                            orders_one,
+                            orders_two,
+                            one_density_matrix,
+                            basis,
+                            points,
+                            transform=transform,
+                            coord_type=coord_type,
+                            deriv_type='general',
+                        )
+                else:
+                    density = evaluate_deriv_reduced_density_matrix(
+                        orders_one,
+                        orders_two,
+                        one_density_matrix,
+                        basis,
+                        points,
+                        transform=transform,
+                        coord_type=coord_type,
+                        deriv_type=deriv_type,
+                    )
                 output += factor * num_occurence * density
     return output
 
--- a/gbasis/evals/density.py
+++ b/gbasis/evals/density.py
@@ -333,6 +333,105 @@ def evaluate_density_gradient(
     return output.T
 
 
+def evaluate_density_laplacian_alt(
+    one_density_matrix, basis, points, transform=None, coord_type="spherical", deriv_type="direct"
+):
+    orders_one_second = np.array(([2, 0, 0], [0, 2, 0], [0, 0, 2]))
+    orders_one_first = np.array(([1, 0, 0], [0, 1, 0], [0, 0, 1]))
+    orders_two = np.array(([1, 0, 0], [0, 1, 0], [0, 0, 1]))
+    output = np.zeros(points.shape[0])
+    # Evaluation of generalized contraction shell for zeroth order = 0,0,0
+    zeroth_deriv = evaluate_deriv_basis(
+        basis, points, np.array([0, 0, 0]),
+        transform=transform, coord_type=coord_type, deriv_type=deriv_type
+    )
+
+    # Evaluation of generalized contraction shell for each partial derivative
+    for orders in orders_one_second:
+        deriv_one = evaluate_deriv_basis(
+            basis, points, orders, transform=transform, coord_type=coord_type, deriv_type=deriv_type
+        )
+
+        density = one_density_matrix.dot(zeroth_deriv)
+        density *= deriv_one
+        output += (2 * 1 * np.sum(density, axis=0))
+
+    for orders in zip(orders_one_first, orders_two):
+        deriv_one = evaluate_deriv_basis(
+            basis, points, orders[0], transform=transform, coord_type=coord_type,
+            deriv_type=deriv_type
+        )
+
+        deriv_two = evaluate_deriv_basis(
+            basis, points, orders[1], transform=transform, coord_type=coord_type,
+            deriv_type=deriv_type
+        )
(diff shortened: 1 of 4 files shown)


gbasis/evals/density.py
```diff
@@ -110,6 +110,7 @@ def evaluate_deriv_reduced_density_matrix(
     points,
     transform=None,
     coord_type="spherical",
+    deriv_type="general"
```
kimt33, line 113: I don't know if `black` is part of QA, but I think it would complain about not having the comma at the end.

gbasis/evals/density.py
```diff
@@ -243,21 +255,34 @@ def evaluate_deriv_density(
...
+                            orders_two,
+                            one_density_matrix,
+                            basis,
+                            points,
+                            transform=transform,
+                            coord_type=coord_type,
+                            deriv_type='general',
+                        )
+                else:
+                    density = evaluate_deriv_reduced_density_matrix(
```
kimt33, line 271: If there are no benefits to using the general method for the orders less than or equal to 2, it may be better to make it the default behaviour.

kimt33: Actually, I change my mind because the `deriv_type` seems to be used quite frequently throughout the code.

kimt33 (dismissed):
Sorry to look at this so late. I haven't been active on the repo but I've been going through some of the PR's just now.

All in all, the PR looks good. It keeps the original functionality so it's easier to test and adds a bunch of performance improvements.

I have two comments though. The first is that it may be better (unless there is a good reason not to) to make the default behavior to use the efficient code if available. This would mostly involve removing the parameter `deriv_type` or checking the orders if available. 

My other comment is that coverage seems to be failing for some reason (for a bunch of the PR's so that might be a repo issue). The tests are not really an issue since they produce the same result as the code before.

Seeing as the repo isn't terribly active, I think it'd be better to merge the PR and leave the nitpicking stuff for someone else as an Issue (rather than PR going stale).

I will approve the PR but will leave it up to current package manager or whoever higher up to make the final call.

kimt33:
Also, the linters QA fails which can be checked with running
`pip install --user tox`
`tox -e qa`
from the package home directory. I think you can get most of them by using `black -l 100` (if not installed `pip install --user black`)

Again, I don't think they're deal-breakers, mostly tedious work that can be passed onto others maybe (or never get done even). Still, I think someone else more involved should make the final call

leila-pujal:
Thanks @kimt33 for your review and sorry for the late reply. I tried to fix all the issues related to "format" I would say and, if I am not wrong, all of them were detected by flake8. Then I have a recurrent issue reported by pylint  `Too many local variables (21/15) (too-many-locals)`. This will take me more time as it means I should look again at the code closely. This is a little bit my fault as I tend to code this way, I don't know if it could be a problem so let me know if you think so and want me to take care.

gbasis/evals/density.py
leila-pujal: It didn't show up using `tox -e qa` but maybe I missed it . My reasoning is that it is the last argument given to the function?

kimt33:
No worries! I don't think there is any rush with this repo. I'm also replying late haha

You don't need to worry about many of the linter errors, especially things like too-many-locals. It normally functions as a discretionary thing to keep some consistency with style. Usually whoever is managing the repository would make the final call or even make corrections, otherwise, it's more or less up to the person making the PR.

Just to be a bit technical, in the cases where PR fails the QA test set in place, you can usually get away with a separate commit which just adds a comment (something like `pylint: disable=too-many-locals`) with an explanation as to why that check was disabled. If the admin has a problem with it, it'd be easier to fix this way.

No need to worry about anything here, since the QA doesn't seem to be working anyways. Thanks again for working on this.

kimt33 (approved):

Issue #119, opened 2021-04-19
[BUG] Base class `BaseTwoIndexSymmetric` in base_two_symm.py can not deal with mix cartesian/spherical coord_types


leila-pujal:
**Describe the bug**
 
I was trying to compute kinetic energy atomic integrals with `gbasis.integrals.kinetic_energy.
kinetic_energy_integral()` with a fchk file where there are mix cartesian and spherical shells (
[neon.fchk.tar.gz](https://github.com/theochem/gbasis/files/6337979/neon.fchk.tar.gz) ) with the code in Reproduce section and got the next error: 

```
File "issue_code.py", line 11, in <module>
    kinetic_integrals = kinetic_energy_integral(basis, coord_type = coord)
  File "/home/leila/gbasis/gbasis/gbasis/integrals/kinetic_energy.py", line 159, in kinetic_energy_integral
    return KineticEnergyIntegral(basis).construct_array_mix(coord_type)
  File "/home/leila/gbasis/gbasis/gbasis/base_two_symm.py", line 304, in construct_array_mix
    cont_one.angmom_components_sph,
  File "/home/leila/gbasis/gbasis/gbasis/wrappers.py", line 117, in angmom_components_sph
    "momentum {0}".format(self.angmom)
ValueError: Given convention does not support spherical contractions for the angular momentum 0
```
If I more or less understood the code correctly in gbasis, this happens because when the code in kinetic_energy.py tries to construct the contractions array with `KineticEnergyIntegral(basis).construct_array_mix(coord_type)` (coord_type = coord in the reproduce code and coord being the list of cartesian/spherical values for each shell) it goes to the base class module in `base_two_symm.py`, to use `construct_array_mix()` function defined there and in this case it builds the transformation variable for all the basis functions https://github.com/theochem/gbasis/blob/c57de8f1cfbe1dbaac8941c7cb9993797c76fca6/gbasis/base_two_symm.py#L301-L326, and after that checks which ones are spherical and applies the transformation. This provokes that the code checks if there is a convention for spherical basis functions in the conventions stored in mol object from IOData but in conventions there is only defined cartesian conventions for angular momentums 0 and 1 because there are the same in both cartesian and spherical. 

In contrast to this behaviour for `BaseTwoIndexSymmetric`, for `BaseOneIndex` in base_one.py this  sorting is done before building the transform variable and avoiding this mistake. 
[...]

Issue #114, opened 2020-07-28
[BUG] Transforms are computed in `construct_array_mix` methods when they are not needed


tovrstra:
**Describe the bug**

The method `construct_array_mix` in `base_two_symm.py`, `base_two_asymm.py` and `base_four.py` constructs `transform_*` variables when they are not needed. When the wrapper to generate the basis set does not define the transformation, this will result in an unwarranted error message. This typically happens when loading wavefunction files with IOData from file formats that always work with Cartesian shells, but there are other cases too, see example below.

A similar problem was fixed in #104 for `base_one.py`.

**To Reproduce**

The following script (with IOData installed) generates the exception:

```python
from importlib.resources import path
from iodata import load_one
from gbasis.wrappers import from_iodata
from gbasis.integrals.overlap import overlap_integral

with path("iodata.test.data", "he2_ghost_psi4_1.0.molden") as fn_full:
        iodata = load_one(str(fn_full))
basis, coord_types = from_iodata(iodata)
olp = overlap_integral(basis, coord_type=coord_types)
print(olp)
```

Exception:

```
Traceback (most recent call last):
  File "example.py", line 9, in <module>
    olp = overlap_integral(basis, coord_type=coord_types)
  File "/home/toon/miniconda3/envs/horton3/lib/python3.7/site-packages/gbasis/integrals/overlap.py", line 145, in overlap_integral
    return Overlap(basis).construct_array_mix(coord_type)
  File "/home/toon/miniconda3/envs/horton3/lib/python3.7/site-packages/gbasis/base_two_symm.py", line 304, in construct_array_mix
    cont_one.angmom_components_sph,
  File "/home/toon/miniconda3/envs/horton3/lib/python3.7/site-packages/gbasis/wrappers.py", line 117, in angmom_components_sph
    "momentum {0}".format(self.angmom)
ValueError: Given convention does not support spherical contractions for the angular momentum 0
```

**Expected behaviour**

```
[[1.00000146e+00 5.95216385e-01 2.59436827e-04 4.41546274e-02]
 [5.95216385e-01 1.00000000e+00 4.41546274e-02 2.14669756e-01]
 [2.59436827e-04 4.41546274e-02 1.00000146e+00 5.95216385e-01]
 [4.41546274e-02 2.14669756e-01 5.95216385e-01 1.00000000e+00]]
```

**Screenshots**

**System Information:**
 - OS: Fedora 32
 - Python version: 3.7.7
 - NumPy version: 1.18.5
[...]

PR #120, opened 2021-07-06
Fix construct_array_mix

Ali-Tehrani:
## Steps

- [✓] Write a good description of what the PR does.
- [✓] Add tests for each unit of code added (e.g. function, class)
- [ ] Update documentation
- [ ] Rebase onto master

## Description
This pull request is regarding the function construct_array_mix within the class BaseTwoIndexSymmetric, BaseTwoIndexAsymmetric, and BaseFourIndexSymmetric. They are further explained in the issue #119 and #114.
In short, the function construct_array_mix transforms from Cartesian to Spherical when it is not necessary, e.g. when one only needs Cartesian form or angular momentum is zero.  This is fixed by adding a if-statement to only apply these transformation when the contraction shell is of type "spherical".

I have added tests suggested in the issue #114 and added my own test with one Cartesian P-type shell and one Spherical D-type shell.  Within the test, the function "construct_array_contraction" is replaced with a simple function that allows one to differentiate between integrating P-P, P-D, D-D type shells, the number of points evaluated is one and the normalization constant are set to one.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Issue #119 and Issue #114

### DIFF
--- a/gbasis/base_four_symm.py
+++ b/gbasis/base_four_symm.py
@@ -435,6 +435,20 @@ class BaseFourIndexSymmetric(BaseGaussianRelatedArray):
             )
         )
         for pair_ind, ((i, cont_one, type_one), (j, cont_two, type_two)) in enumerate(pair_i_cont):
+            if type_one == "spherical":
+                transform_one = generate_transformation(
+                    cont_one.angmom,
+                    cont_one.angmom_components_cart,
+                    cont_one.angmom_components_sph,
+                    "left",
+                )
+            if type_two == "spherical":
+                transform_two = generate_transformation(
+                    cont_two.angmom,
+                    cont_two.angmom_components_cart,
+                    cont_two.angmom_components_sph,
+                    "left",
+                )
             for (k, cont_three, type_three), (l, cont_four, type_four) in pair_i_cont[pair_ind:]:
                 block = self.construct_array_contraction(
                     cont_one, cont_two, cont_three, cont_four, **kwargs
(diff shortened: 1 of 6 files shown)


gbasis/base_four_symm.py
```diff
@@ -475,15 +450,32 @@ def construct_array_mix(self, coord_types, **kwargs):
                 block *= cont_four.norm_cont.reshape(
                     1, 1, 1, 1, 1, 1, *block.shape[6:8], *[1 for _ in block.shape[8:]]
                 )
-
                 # transform
                 if type_one == "spherical":
+                    transform_one = generate_transformation(
```
kimt33: Just to comment for context:
I think the reason these were defined before was to avoid recomputing the `transform_one` and `transform_two` for each iteration of the for loop over `k` and `l`.

I don't think it makes any performance difference (because this is not the bottleneck of the code) and may be it's better to test it here especially if it introduces bugs.

kimt33 (dismissed):
Looks good

I don't know if the QA stopped working, but I can't seem to see the QA results on travis for this PR for some reason. But either way, it's a bug fix and it looks fine to me.

gbasis/base_four_symm.py
Ali-Tehrani: Thanks David, you're right and I added these back into the loop-body to avoid recomputing them.  I don't know why QA stopped working but I ran tox locally and no issues with linters or tests/coverage were found.

kimt33 (dismissed):
Looks good

Ali-Tehrani:
Thanks, I'll just merge this now.  I squashed the commits and did a force-push here.

PR #135, opened 2023-10-26
Website

RichRick1:
Build a website for the package.

Documentation, the website built and deployment should be done automatically with every push on the main branch

@leila-pujal  There are few things to be done:
1. Change the deployment branch to gh-pages in the settings
2. Change the default host name to gbasis.qcdevs.org
3. Provide more examples of usage the code, once we have them
Also @leila-pujal  @Ali-Tehrani @PaulWAyers @FarnazH  any comments/suggestions are very welcomed


FarnazH (commented):
@RichRick1, thanks for your PR. I have a few questions:

1. Do the `rst` files like `website/_autosummary/gbasis.rst` need to committed? I remember that they were automatically generated by packages like Sphinx. In that case, when the code/API changes, these `rst` files didn't need to be updated.
2. The same goes with `html` files `website/_build/html/_autosummary/gbasis.html`.

For example, see https://github.com/QC-Devs/chemtools/tree/master/doc. Please let me know how what you are doing is different.

Ali-Tehrani:
I'm currently working on this pull-request too and need to submit a few commits to it.

1. The .rst autosummary is generated using sphinx-apidoc.  So currently the user has to run the command
`sphinx-apidoc --separate -o website/_autosummary gbasis -M -f` 
2. The html is generated as well using jupyter-book

I'm removing both _autosummary and the _build (including the html) folder and instead I'm writing a README.md on how to generate the documentation locally.  Rick's pull-request that automatically builds and uploads to the website should handle the rest.

RichRick1:
I run the `sphinx-apidoc -o website/_autosummary gbasis -M -f` in the GitHub actions. I think it should do the similiar as `sphinx-apidoc --separate -o website/_autosummary gbasis -M -f` the only difference is it produces fewer number of files. 
So, there is no need to commit those files (same with html). Thanks for catching this

Ali-Tehrani:
The website looks good to me.  I've added the README on how to build documentation and push to website, fixed up latex formulas not rendering, fixed an issue when the method arguments goes out of page and removed both `_autosummary` and `_build`.  I have not checked whether the formulas are correct or not

Ali-Tehrani (approved):

RichRick1:
@Ali-Tehrani Thanks! Also, I don't think we need the README file, because in the GitHub actions website is build and deployed automatically

RichRick1:
All the deployment and building the documentation with sphinx is here https://github.com/RichRick1/gbasis/blob/0c2f312443c1214391e024259b7eddfd1238ac4b/.github/workflows/main.yml#L47

FarnazH (approved):
Thanks @Ali-Tehrani and @RichRick1. It is good to keep the README.md (for now, at least) to provide more instructions for making the website, as not everyone is comfortable with how the GitHub pages work.

Issue #117, opened 2021-03-08
[BUG] Negative Electron Density Values


FarnazH:
**Describe the bug**
I was integrating density^(1/3) when I noticed that the electron density computed by `evaluate_density` gives a negative, but very small value, for faraway points (e.g. `-2.77765871e-128`). 

**To Reproduce**
Helium FCHK file: [helium.fchk.tar.gz](https://github.com/theochem/gbasis/files/6104600/helium.fchk.tar.gz)

```python
import numpy as np

from iodata import load_one

from grid.molgrid import MolGrid
from grid.becke import BeckeWeights
from grid.rtransform import BeckeTF
from grid.onedgrid import GaussChebyshevType2

from gbasis.wrappers import from_iodata
from gbasis.evals.density import evaluate_density

# load molecule
mol = load_one("helium.fchk")
dm = mol.one_rdms['scf']
print("molecular charge   = ", mol.charge)
print("atomic numbers     = ", mol.atnums)
print("atomic coordinates = ", mol.atcoords)

# make grid
oned = GaussChebyshevType2(150)
rgrid = BeckeTF(1.0e-6, 1.5).transform_1d_grid(oned)
grid = MolGrid.horton_molgrid(mol.atcoords, mol.atnums, rgrid, 350, BeckeWeights(order=3))

# evaluate density
basis, coord_types = from_iodata(mol)
dens = evaluate_density(dm, basis, grid.points, coord_type=coord_types)
print("min & max density = ", np.min(dens), np.max(dens))

index = np.where(dens < 0.0)
print("density     at points with negative densiry = ", dens[index])
print("coordinates of points with negative density = ", grid.points[index])
```

The output is:

```bash
molecular charge   =  0.0
atomic numbers     =  [2]
atomic coordinates =  [[0. 0. 0.]]
min & max density =  -2.77765870781763e-128 2.930734188084798
density     at points with negative densiry =  [-2.77765871e-128]
coordinates of points with negative density =  [[ 10.15753086 -15.57090172 -10.15753086]]
```

**Expected behaviour**
The electron density is expected to be positive at every point.

**System Information:**
 - OS: macOS
 - Python version: 3.6.10
 - NumPy version: 1.18.1
 - SciPy version: 1.4.1

**Additional context**
I am using the snippet below as a workaround, until this get fixed.
`dens[np.where((dens < 0.0) & (abs(dens) < 1.0e-15))] = 0.0`

PaulWAyers:
[...]

PR #136, opened 2023-11-01
Clip Negative Electron Density Values from `evaluate_density`

maximilianvz:
## Steps

- [ ] ~Add tests for each unit of code added (e.g. function, class)~ (this didn't require new tests)
- [X] Update documentation
- [ ] Rebase onto master

## Description

In accordance with @FarnazH's request in Issue #117, I have used [Numpy's clip method](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.clip.html) in the statement that returns the density array in  `evaluate_density`. By specifying `.clip(min=0.0)`, negative values in the array will be reassigned to a value of 0.0 (this is done in-place).

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/gbasis/evals/density.py
+++ b/gbasis/evals/density.py
@@ -99,7 +99,7 @@ def evaluate_density(one_density_matrix, basis, points, transform=None, coord_ty
 
     """
     orb_eval = evaluate_basis(basis, points, transform=transform, coord_type=coord_type)
-    # use numpy.ndarray.clip method in return to remove small negative values
+    # Fix: # 117; to avoid small negative density values, the array is clipped
     return evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval).clip(min=0.0)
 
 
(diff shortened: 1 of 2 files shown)


gbasis/evals/density.py
```diff
@@ -99,7 +99,8 @@ def evaluate_density(one_density_matrix, basis, points, transform=None, coord_ty
 
     """
     orb_eval = evaluate_basis(basis, points, transform=transform, coord_type=coord_type)
-    return evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval)
+    # use numpy.ndarray.clip method in return to remove small negative values
+    return evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval).clip(min=0.0)
```
FarnazH: @maximilianvz, can you please add a oneline comment above the return statement like "Fix: https://github.com/theochem/gbasis/issues/117; to avoid small negative density values, the array is clipped."

FarnazH (changes requested):
@maximilianvz, thanks for your PR. There are already tests computing the density, can you add a check for those to test that the density is positive at all points. I wonder whether we would have caught this earlier if we had tests like that.

FarnazH:
@maximilianvz, some tests fail in this branch, so can you rebase your branch on the latest `master` and force push to `fix_issue_117`?

FarnazH (approved):
Thanks @maximilianvz. Ready to be merged, and close issue #117.

PR #138, opened 2023-11-17
Fix factorial2

### LINKED ISSUE #129: Codes using `factorial2` function
**Describe the bug**
The `factorial2` function in `SciPy` returns different values for negative inputs depending on the version. For versions older than `1.11.0`, it returns 1, while for versions `1.11.0` and newer, it returns 0.

Code that utilizes the `factorial2` function may encounter potential issues, such as those found in the `contractions.py` file:

```python
    @property
    def norm_prim_cart(self):
        r"""Return the normalization constants of the Cartesian Gaussian primitives.

        For a Cartesian primitive with exponent :math:`\alpha_i`, the normalization constant is:

        .. math::
           N(\alpha_i, \vec{a}) = \sqrt {
           \left(\frac{2\alpha_i}{\pi}\right)^\frac{3}{2}
           \frac{(4\alpha_i)^{a_x + a_y + a_z}}{(2a_x - 1)!! (2a_y - 1)!! (2a_z - 1)!!}}

        Returns
        -------
        norm_prim_cart : np.ndarray(L, K)
            The normalization constants of the Cartesian Gaussian primitives.
            `L` is the number of contracted Cartesian Gaussian functions for the given angular
            momentum, i.e. :math:`(\ell + 1) * (\ell + 2) / 2`
            `K` is the number of exponents (i.e. primitives).

        """
        exponents = self.exps[np.newaxis, :]
        angmom_components_cart = self.angmom_components_cart[:, :, np.newaxis]

        return (
            (2 * exponents / np.pi) ** (3 / 4)
            * ((4 * exponents) ** (self.angmom / 2))
            / np.sqrt(np.prod(factorial2(2 * angmom_components_cart - 1), axis=1))
        )
```

An unwanted value will be returned when `angmom_components_cart` is 0, i.e., $s$ orbitals.

**To Reproduce**
Use `SciPy` with version $\geq 1.11.0$

FarnazH:
## Steps

- [x] Update documentation
- [x] Rebase onto master

## Description
This PR addresses issue #129 by adding a `factorial2` function that wraps `scipy.special.factorial2` to return 1.0 when the input is not positive. This is temporary additional, while we wait for Scipy's update. To learn more, see https://github.com/scipy/scipy/issues/18409. This function is added in `gbasis.utils` (I tried adding it to `gbasis.wrappers` first but that would cause circular imports).

Just patching scipy version (as was done in f74123ebf4b640333bf1a3aec4460b1766a7d046) was not enough, because unless you install the `gbasis` package, the correct version of `scipy` is not used causing headache. Also, one might need to use scipy `1.11.x` in their environment for other purposes. In the meanwhile, correcting the `factorial2` output seems to me a more practical solution than patching the scipy version.

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes #129

### DIFF
--- a/gbasis/contractions.py
+++ b/gbasis/contractions.py
@@ -1,6 +1,6 @@
 """Data class for contractions of Gaussian-type primitives."""
+from gbasis.utils import factorial2
 import numpy as np
-from scipy.special import factorial2
 
 
 class GeneralizedContractionShell:
--- a/gbasis/integrals/_one_elec_int.py
+++ b/gbasis/integrals/_one_elec_int.py
@@ -1,6 +1,6 @@
 """One-electron integrals involving Contracted Cartesian Gaussians."""
 import numpy as np
-from scipy.special import factorial2
+from gbasis.utils import factorial2
 
 
 # pylint: disable=C0103,R0914,R0915
--- a/gbasis/integrals/_two_elec_int.py
+++ b/gbasis/integrals/_two_elec_int.py
@@ -1,6 +1,6 @@
 """Two-electron integrals involving Contracted Cartesian Gaussians."""
 import numpy as np
-from scipy.special import factorial2
+from gbasis.utils import factorial2
 
 # pylint: disable=C0103,R0914,R0915
 
--- a/gbasis/spherical.py
+++ b/gbasis/spherical.py
@@ -3,7 +3,8 @@
 from collections import defaultdict
 
 import numpy as np
-from scipy.special import comb, factorial, factorial2
+from scipy.special import comb, factorial
+from gbasis.utils import factorial2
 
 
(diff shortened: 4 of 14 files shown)


Ali-Tehrani (approved):

PaulWAyers (approved):
Looks good to me. Thanks @FarnazH 

Are there any other pieces of `gbasis` that really belong in `utils.py`? I can't think of any right now, but probably eventually some other things (e.g., functions related to the Boys function) belong there.

FarnazH:
@PaulWAyers, I cannot think of any other function right now, but it is good to keep in mind.

leila-pujal (approved):

Issue #113, opened 2020-07-10
[REQUEST] Merging coord_type into GeneralizedContractionShell


tovrstra:
**Is your feature request related to a problem? Please describe.**

This came up when I was tuning the `from_iodata` function in `gbasis/wrappers.py`. At the moment, this function has two return values: `basis`, which is a list of `GeneralizedContractionShell` instances and `coord_types`, which says for each shell whether it is Cartesian or pure. I would suggest to make the elements of coord_types attributes of the corresponding GeneralizedContractionShell instances. (This can be done in various was effectively, not necessarily literally the way I wrote.)

The motivation for this request is that the `coord_types` part is in practice always tied to the rest of the definition of the basis set. (It is normally not something that is varied during a calculation.) At the moment, when using GBasis, one has to keep track of both variables and pass then into the eval functions every time, which seems redundant and error prone. Forgetting to pass in `coord_types` would essentially always be a bug.

**Is there a specific integral/formula that you would like implemented?**

No

**Is there a change to the code or algorithm you would like to see?**

To be discussed. I can look into options to streamline both variables into one. My first impression is that one could, at the same timel simplify and reduce a lot of the code in the `base*.py` modules.

**Additional context**

None

maximilianvz:
I need to leave a comment on this issue in order to be assigned to it by @leila-pujal.

leila-pujal:
@maximilianvz  and I have looked at ways to incorporate this change. The `coord_type` variable at the moment is only returned from `from_iodata` but in theory, you can specify the coordinate type for any list of `GeneralizedContractionShell` class objects list. 

Gbasis expects the user to provide information about the coordinate type of a particular shell being defined in cartesian or spherical coordinates. It can either be specified as a whole by just passing "cartesian" or "spherical" strings or shell by shell in a list of strings being either "cartesian" or "spherical". This is a check for I think any evaluation or integral function defined in each module (.py file) to basically sort what kind of contraction_array_* function defined in base_one , base_two and base_four modules to choose.  An example of this for the overlap:
https://github.com/theochem/gbasis/blob/5fcd1904475e7cb0b4c4a0d83c1f980de1b42bf7/gbasis/integrals/overlap.py#L137-L145

Below I explain 2 options that we have thought of: 
[...]

PR #140, opened 2023-11-22
Merging `coord_type` into GeneralizedContractionShell

maximilianvz:
## Steps

- [X] Add tests for each unit of code added (e.g. function, class)
- [X] Update documentation
- [ ] Rebase onto master

## Description

This pull request addresses what was requested in Issue #113. Thus far, I have pushed 6 commits to my forked repo, which I will describe below:

[Add coord_type attr to GeneralizedContractionShell](https://github.com/theochem/gbasis/commit/fa125cc2cb094ae4b8a0829b7e51a7e807fce269)

This is perhaps the most fundamental commit in the PR. The commit message describes it pretty well: I have added the `coord_type` attribute to the GeneralizedContractionShell class. When creating an instance of the GeneralizedContractionShell class, one should pass a string that describes the coordinate type of the contraction. This string should either be "c"/"cartesian" to specify the *cartesian* coordinate type, or "p" (pure)/"spherical" to specify the *spherical* coordinate type. By default, "p" is provided, so the spherical coordinate type is assumed if nothing is passed for instantiation.

[Negate from_iodata return of coord_types](https://github.com/theochem/gbasis/commit/71cda1e1b12eb1c68e00f185dc5991441fd7826f)

With the addition of the `coord_types` attribute to GeneralizedContractionShell objects, there is no need for `from_iodata` to return `coord_types`, as the `coord_types` get embedded in the `basis` that the function returns. Of course, I needed to pass the `coord_types` to the `IODataShell` objects in this function to ensure my last point stands.

[Update make_contractions](https://github.com/theochem/gbasis/commit/02919109627e19e4af47670fc0275e3e288afb6a)

The [`make_contractions`](https://github.com/theochem/gbasis/blob/5fcd1904475e7cb0b4c4a0d83c1f980de1b42bf7/gbasis/parsers.py#L168) function is useful for instantiating a basis (list of GeneralizedShellContraction objects) for a system of atoms based on the `basis_dict` output of the parsers from [`gbasis.parsers`](https://github.com/theochem/gbasis/blob/5fcd1904475e7cb0b4c4a0d83c1f980de1b42bf7/gbasis/parsers.py). Previously, it only received the `basis_dict`, atoms at which the contractions are centered, and the coordinates of each atom. However, if we are to embed `coord_types` into the basis objects that we create, `make_contractions` must be able to handle that as an input as well. I have set this up such that, for a given call of `make_contractions`, one can pass either a string ("cartesian" or "spherical") or a list of strings to specify `coord_types`. If a string is passed, **all** contractions for **all** atoms are treated as having the same `coord_type`. However, if a list is passed, it must be the same length as the total number of contractions across all atoms. For example, if atom 1 contributes 4 contractions and atom 2 contributes 5 contractions, the `coord_types` list should be of length 9, where the first 4 elements correspond to contractions for atom 1 and the last 5 elements correspond to contractions for atom 2.

As I am writing this, I realize that maybe a user could try to pass "c" or "p" as a string in place of `coord_types`. Likely, they would do this expecting it to assign cartesian or spherical coordinate types to each contraction, respectively, though this would raise an error. This begs the question of whether I should change this commit by changing `if coord_types == "spherical":` to `if coord_types in ["spherical", "p"]:`, for example, or to change the allowed arguments for instantiation in the GeneralizedContractionShell class (this option requires me to alter my first commit described above).

[Remove coord_type args from evals/integrals](https://github.com/theochem/gbasis/commit/92ba2dbab8444882d042ae4a353f85e333a9683f)

This is the meat of where I had to deal with the fact that we no longer want to have to pass `coord_types` for eval/integral functions in our API. Essentially, if we embed `coord_types` in our `basis` variables (which are lists of GeneralizedContractionShell objects, now containing `coord_type` attributes), we should only have to pass `basis` to these functions, and then extract the `coord_types` from the contractions so that they can be passed into the `construct_array*` functions in the `base*.py` modules. Generally, I do this via `coord_type = [type for type in [shell.coord_type for shell in basis]]`. This is done internally, so the user need not worry about passing `coord_types` to these functions. In fact, if they try to pass `coord_types`, they'll encounter an error, as I removed the optional `coord_type` arguments from the eval/integral functions. 

[Update docs and coord_type checks in base*](https://github.com/theochem/gbasis/commit/49e1b18dd9e664d5cf9456d636318812e0366ce4)

Because I extract `coord_type` as a list prior to passing into the `contruct_array*` functions of the `base*.py` modules, `construct_array_cartesian` and `construct_array_spherical` don't actually get used based on the pre-existing checks. Previously, these functions would only be used if `coord_types == "cartesian"` or `coord_types == "spherical"`, but if `coord_types` is a list, this obviously won't check out. Thus, to address this, I added checks that, for example, `all(type == "cartesian" for type in coord_type)` to see if `construct_array_cartesian` can be used. This raises an important point that I'd like to be addressed:

With `coord_types` being passed as a list, I *could* get away with just using `construct_array_mix` across the board (i.e., all the tests will pass). I don't know if there is a boost in efficiency that can be achieved by using `construct_array_cartesian/spherical` as opposed to `construct_array_mix`, but if there is **not**, I bet we could simplify the `base*.py` modules a lot by removing `construct_array_cartesian/spherical` and replacing `construct_array_mix` with `construct_array`, which would be able to handle homogeneous or heterogeneous `coord_type` lists equally well. I do realize that `construct_array_lincomb` makes calls to `construct_array_cartesian/spherical` though, so things would be a bit more complicated than what I've described. I didn't want to go that far in this PR, in case there's a benefit to maintaining the use of `construct_array_cartesian/spherical`, but it was an important point that I wanted to bring up in any case.

[Adjust tests for coord_types API change](https://github.com/theochem/gbasis/commit/a5670a98414109922036c63c21116e7828967f8b)

This one should be pretty self-explanatory. When I removed the `coord_type` optional arguments from the evals/integrals modules, any tests that called these functions would cause an error. In general, I just modified tests so that they didn't use coord_types as an argument for evals/integrals functions, but instead specified `coord_types` when creating a `basis` variable, which then gets passed to evals/integrals functionality.

One place where I sort of brute-forced the tests to work was when `construct_array_lincomb` was tested directly against some other functionality. Here's an example of what I mean:

Screenshot 2023-11-22 at 4 41 07 PM

You'll notice that at the bottom of that screenshot, I changed `angular_momentum_integral_obj.construct_array_lincomb(transform, "spherical"),` to `angular_momentum_integral_obj.construct_array_lincomb(transform, ["spherical"]),`. The change here is that `"spherical"` became `["spherical"]` (I passed a list instead of a string). I call this "brute-forced" because, based on my changes in checks of `coord_types` in `base*.py` modules (see the description of the commit directly above this one), `coord_types` should be received in those modules as a list. When calling `make_contractions`, as I described above, you **don't** have to pass `coord_types` as a list - if you pass a string it'll work and treat all contractions with the same coordinate type. Based on that, one might want to pass `coord_types` to `construct_array_lincomb` as a string in tests. However, I don't believe users should be directly accessing functions in the `base*.py` modules, so only using this "brute-forced" change of passing a list versus a string should hopefully be okay just in the context of testing.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -138,13 +138,19 @@ def from_iodata(mol):
     return basis
 
 
-def from_pyscf(mol):
+def from_pyscf(mol, coord_types="spherical"):
     """Return basis set stored within the `Mole` instance in `pyscf`.
 
     Parameters
     ----------
     mol : pyscf.gto.mole.Mole
         `Mole` object in `pyscf`.
+    coord_types : {"cartesian", list/tuple of "cartesian" or "spherical", "spherical"}
+        Types of the coordinate system for the contractions.
+        If "cartesian", then all of the contractions are treated as Cartesian contractions.
+        If "spherical", then all of the contractions are treated as spherical contractions.
+        If list/tuple, then each entry must be a "cartesian" or "spherical" to specify the
+        coordinate type of each `PyscfShell` instance.
 
     Returns
     -------
@@ -183,6 +189,7 @@ def from_pyscf(mol):
             return super().angmom_components_sph
 
     basis = []
+    len_coord_types = len(coord_types)
     for atom, coord in mol._atom:
         basis_info = mol._basis[atom]
 
@@ -194,6 +201,22 @@ def from_pyscf(mol):
 
             coeffs = exps_coeffs[:, 1:]
 
-            basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps)))
+            if type(coord_types) == str:
+            # if coord_types given as a single string, assign the specified type to all contractions for all atoms
+                if coord_types == "spherical":
+                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), 'p'))
+                elif coord_types == "cartesian":
+                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), 'c'))
+                else:
+                    raise ValueError("If coord_types is a string, it must be either 'spherical' or 'cartesian'.")
+            elif type(coord_types) == list:
+                # if coord_types given as a list, assign the specified type to each atom's contractions individually
+                if len_coord_types == mol.nbas:
+                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), coord_types.pop(0)))
+                else:
+                    raise ValueError(
+                        "If coord_types is a list, it must be the same length as the total number of contractions.")
+            else:
+                raise TypeError("coord_types must be a string or list of strings.")
 
     return tuple(basis)
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -138,19 +138,13 @@ def from_iodata(mol):
     return basis
 
 
-def from_pyscf(mol, coord_types="spherical"):
+def from_pyscf(mol):
     """Return basis set stored within the `Mole` instance in `pyscf`.
 
     Parameters
     ----------
     mol : pyscf.gto.mole.Mole
         `Mole` object in `pyscf`.
-    coord_types : {"cartesian", list/tuple of "cartesian" or "spherical", "spherical"}
-        Types of the coordinate system for the contractions.
-        If "cartesian", then all of the contractions are treated as Cartesian contractions.
-        If "spherical", then all of the contractions are treated as spherical contractions.
-        If list/tuple, then each entry must be a "cartesian" or "spherical" to specify the
-        coordinate type of each `PyscfShell` instance.
 
     Returns
     -------
@@ -173,6 +167,10 @@ def from_pyscf(mol, coord_types="spherical"):
     # pylint: disable=W0212
     if not (mol.__class__.__name__ == "Mole" and hasattr(mol, "_basis")):
         raise ValueError("`mol` must be a `pyscf.gto.mole.Mole` instance.")
+
+    # assign the coordinate types (which can be either Cartesian or Spherical)
+    # it seems like pyscf does not support mixed "cartesian" and "spherical" basis.
+    coord_types = "cartesian" if mol.cart else "spherical"
 
     class PyscfShell(GeneralizedContractionShell):
         """Shell object that is compatible with gbasis' shell object.
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -132,7 +132,9 @@ def from_iodata(mol):
 
         # pylint: disable=E1136
         basis.append(
-            IODataShell(angmom, mol.atcoords[shell.icenter], shell.coeffs, shell.exponents, shell.kinds[0])
+            IODataShell(
+                angmom, mol.atcoords[shell.icenter], shell.coeffs, shell.exponents, shell.kinds[0]
+            )
         )
 
     return basis
@@ -187,34 +189,14 @@ def from_pyscf(mol):
             return super().angmom_components_sph
 
     basis = []
-    len_coord_types = len(coord_types)
     for atom, coord in mol._atom:
         basis_info = mol._basis[atom]
 
         for shell in basis_info:
             angmom = shell[0]
-
             exps_coeffs = np.vstack(shell[1:])
-            exps = exps_coeffs[:, 0]
-
-            coeffs = exps_coeffs[:, 1:]
-
-            if type(coord_types) == str:
-            # if coord_types given as a single string, assign the specified type to all contractions for all atoms
-                if coord_types == "spherical":
-                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), 'p'))
-                elif coord_types == "cartesian":
-                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), 'c'))
-                else:
-                    raise ValueError("If coord_types is a string, it must be either 'spherical' or 'cartesian'.")
-            elif type(coord_types) == list:
-                # if coord_types given as a list, assign the specified type to each atom's contractions individually
-                if len_coord_types == mol.nbas:
-                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), coord_types.pop(0)))
-                else:
-                    raise ValueError(
-                        "If coord_types is a list, it must be the same length as the total number of contractions.")
-            else:
-                raise TypeError("coord_types must be a string or list of strings.")
+            exps = np.array(exps_coeffs[:, 0])
+            coeffs = np.array(exps_coeffs[:, 1:])
+            basis.append(PyscfShell(angmom, np.array(coord), coeffs, exps, coord_types))
 
     return tuple(basis)
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -16,11 +16,6 @@ def from_iodata(mol):
     basis : tuple of gbasis.contraciton.GeneralizedContractionShell
         Basis set object used within the `gbasis` module.
         `GeneralizedContractionShell` corresponds to the `Shell` object within `iodata.basis`.
-    coord_types : list of str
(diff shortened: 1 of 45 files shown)


FarnazH (changes requested):
@maximilianvz I noticed that ` tests/test_eval.py::test_evaluate_basis_pyscf` fails. I removed the argument `coord_type` in 5a52509, but it still failed because the `from_pyscf` uses spherical `coord_type` by default. Can you fix this?

This makes me also wonder whether the `coord_type` argument of `GeneralizedContractionShell` should be an optional argument (instead of being required). Did you choose the default value of `“p”` because the `evaluate*` functions had such defaults? 

Thanks for your detailed description in this PR. I appreciate the effort you have put into explaining the changes you have made. For the most part, you don’t need to explain in words the changes you made to the code. The reviewer is familiar with the code, and considering that your commits are bit-sized, it should be clear to the reviewer what has happened. This can save you a lot of time in making a PR. For example, looking at [Adjust tests for coord_types API change](https://github.com/theochem/gbasis/commit/a5670a98414109922036c63c21116e7828967f8b), the changes you have made and why you have made them is clear because you have made a good commit fixing one thing. If anything is unclear the reviewer can ask. I recommend only explaining your logic when it is not obvious from the code.

Instead, it is more useful to add these descriptions to the
1. body of your commit messages (leave an empty line after your commit message/title and then include a body), because then your description stays with the commit and is easier to refer to (than finding the PR where the changes were made). Like 5a52509a1ec4cf32bfe79573439aadd5489825f0

2. Or, sometimes to the docstring. See my commit 6109e8d850a2e41e025cadaf49d7ab06ce3349b0.

In QC-Devs packages, we try not to provide the default value in the docstring. See 6109e8d850a2e41e025cadaf49d7ab06ce3349b0. The default is visible in `def ...` line. This makes the doctoring shorter and easier to maintain (it can easily go unnoticed if someone changes the default value but don’t change the docstring).

In [Remove coord_type args from evals/integrals](https://github.com/theochem/gbasis/commit/92ba2dbab8444882d042ae4a353f85e333a9683f), 
`type` is a built-in function name, so it should not be used as a variable name. your editor should color code that. Change the variable name (e.g., to ct). 
What if `coord_type` is “c” or “p”?

maximilianvz:
@FarnazH, I believe I've addressed both of the changes you asked for:

1. I got `tests/test_eval.py::test_evaluate_basis_pyscf` to pass by requiring that the user pass the coordinate types of shells when making a `basis` list via `from_pyscf`. Using `from_iodata`, there was an [elegant way to extract this information from `mol`](https://github.com/maximilianvz/gbasis/blob/d6a35db836b3e42dcb9f672de8ac31607d1f5c1e/gbasis/wrappers.py#L135C93-L135C107) (such that the user didn't have to specify `coord_types` in the call to the function), but I don't know if this is possible with `from_pyscf`, so I went with this workaround.
2. I stopped using the reserved `type` keyword inappropriately, as you pointed out.

Addressing your comments:

1. I doubt it would be very difficult to make `coord_type` a required argument for `GeneralizedContractionShell`. You are correct in guessing that I used a default value of `p` because the `eval*` functions were doing this previously.
2. In general, if the user is using `make_contractions` or `from_pyscf` to build a `basis` list, they should only pass "spherical", "cartesian", or a list of these strings (actually, now that I write this, I realize you could pass a list of "c"s and "p"s, which slipped under my radar). Passing "c" or "p" individually is technically allowed if you're accessing `GeneralizedContractionShell` directly, but these higher-level functions don't recognize those as valid inputs. Evidently, there's an inconsistency here. I can address this by including the possibility of passing `c` and `p` as single strings to `make_contractions` and `from_pyscf`. Let me know if you have other ideas.

FarnazH:
Thanks @maximilianvz. Based on your comments, 

1. Let's make `coord_type` a required argument for `GeneralizedContractionShell`.
2. Let's make sure higher level functions support `'c'` and `'p'` as well.

gbasis/wrappers.py
```diff
@@ -203,6 +199,22 @@ def angmom_components_sph(self):
...
+                if coord_types == "spherical":
+                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), 'p'))
+                elif coord_types == "cartesian":
+                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), 'c'))
+                else:
+                    raise ValueError("If coord_types is a string, it must be either 'spherical' or 'cartesian'.")
+            elif type(coord_types) == list:
+                # if coord_types given as a list, assign the specified type to each atom's contractions individually
+                if len_coord_types == mol.nbas:
+                    basis.append(PyscfShell(angmom, np.array(coord), np.array(coeffs), np.array(exps), coord_types.pop(0)))
```
FarnazH: @maximilianvz, the problem with popping elements from `coord_types` list is that, if there is more than 1 atom in the molecule, the `coord_types` list will be empty after the shells for the first atom are appended to `basis`. This tentative bug was not caught by the existing tests, because they all had a string `coord_type`.

Considering the recent change in 729d83e92d19e87b545e5aff64b575639ea449b2, and the fact that PySCF does not support mixed coordinate types (based on my quick look), testing this can be tricky (any thoughts?). Then, we should either only support `str` type `coord_types` or index the `coord_types` list (instead of popping). The latter is less desitable, because I am not still sure how to test a list `coord_types`.

@leila-pujal, @RichRick1, @msricher, and @PaulWAyers, do you know whether PySCF supports mixed "cartesian" and "spherical" basis?

leila-pujal: From what I remember using PySCF, it does not allow for mixed cartesian/spherical. There's an attribute in the `Mole` class called `cart` that's boolean. `True` the basis is cartesian and `False` spherical.

maximilianvz: @FarnazH, if PySCF doesn't allow for mixed cartesian/spherical coordinates, then we should remove the `elif type(coord_types) == list:` clause altogether, as the only time you would want to pass a list here is if you were using a mix of these coordinate systems. By adding [this line](https://github.com/maximilianvz/gbasis/blob/610dd8f8d9777244ba7a7836293cf9762793a30e/gbasis/wrappers.py#L173), you prevent `coord_types` from being a list anyway. Thus, I don't see any "bug" applying to `from_pyscf` if we were to remove that part of the code. 

However, [`make_contractions`](https://github.com/maximilianvz/gbasis/blob/610dd8f8d9777244ba7a7836293cf9762793a30e/gbasis/parsers.py#L168) does allow for mixed coordinate systems. It also uses this `elif type(coord_types) == list:` clause, but I believe [this test](https://github.com/maximilianvz/gbasis/blob/610dd8f8d9777244ba7a7836293cf9762793a30e/tests/test_overlap_asymm.py#L58C8-L58C8) shows that popping from the `coord_types` list will work. In that example of a Krypton dimer, each Krypton atom contributes 5 shells, for a total of 10 shells. In the call to `make_contractions`, a list of length 10, where all but the last entries are "spherical", is passed. [You will only ever `pop` from `coord_types` if the initial length of `coord_types` is equal to the TOTAL number of contractions in the molecule](https://github.com/maximilianvz/gbasis/blob/610dd8f8d9777244ba7a7836293cf9762793a30e/gbasis/parsers.py#L225C17-L225C17). In this case, the onus is on the user to specify a list of length 10 to `make_contractions`, and to ensure that they know that the first 5 coordinate types in this list would get applied to the 5 shells of the first Krypton atom, while the last 5 coordinate types in that list (where the last element was specified as "cartesian") would get applied to the five shells of the second Krypton atom.  

Please let me know if you have any questions or doubts.

FarnazH: @maximilianvz, based on our discussions, I made the changes below. I believe this is good to be merged, let me know if you have any comments.

Take a look at the commits on how the code is refactored to avoid nested `if/elif/else` statements. Also, when checking the type of a variable, it is more appropriate to use `isinstance` function.

FarnazH (approved):

Issue #116, opened 2021-01-07
[DOCS] Wrappers documentation needs an update


wilhadams:
When #106 was pulled in, I didn't notice that we should update the notes.pdf file as well, because the wrappers return both the `basis` and, newly, `coord_types`. I'll try to update the docs soon.

maximilianvz:
Apparently, in order to be assigned to issues in this repository, I need to leave a comment somewhere. This is the point of this comment.

gabrielasd:
Hi @leila-pujal, just to make sure, the notes under the [docs](https://github.com/theochem/gbasis/tree/master/docs) folder are the latest version?

leila-pujal:
Hi @gabrielasd, yes those notes are the latest. But now I was thinking @maximilianvz is working on issue #113 and basically the newly added `coord_typ` variable that was returned we are planning to incorporate that as an attribute of the `GeneralizedContractionShell` class. I think you can start updating the notes incorporating all the modifications coming from IOData incorporated in PR #106 except the `coord_type` . There should be a draft for issue #113 soon so then you can update the notes based on that.  Let me know what you think.

gabrielasd:
OK! Yes, that makes sense. thanks @leila-pujal

PaulWAyers:
There are also (more scientific, less technical) notes in the Discussion thread in case that is helpful.
https://github.com/orgs/theochem/discussions/2

gabrielasd:
> There are also (more scientific, less technical) notes in the Discussion thread in case that is helpful. https://github.com/orgs/theochem/discussions/2

@PaulWAyers I could not access this link. I can get to theochem, but I don't see the discussion board there.

gabrielasd:
@leila-pujal 
For the docstrings in the wrappers module, I've done a 1st read and update in my fork of the project. They were minor changes in the `from_iodata` function. I left out any modifications dealing with the `coord_type` variable like we talked and I've seen that @maximilianvz already updated this part within his PR.
For the update in the latex notes, I've only been looking at section 4 "Basis Set" since it seemed to me the most related to this issue. However, should I check other sections too?

Also, should I include the examples for how to use `from_iodata` and `from_pyscf` from the notes as part of the wrappers docstrings?

PR #141, opened 2023-12-14
Wrappers docs issue 116

gabrielasd:
## Steps

- [ ] Write a good description of what the PR does.
- [x] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |

## Related Issue
This PR addresses issue #116 requesting the update of the docs/notes.pdf following changes made to the IOData wrapper in PR #106. 

Changes:
- Update `from_iodata` docstrings in the wrappers module removing no-longer-existent raised errors.
- Update section 4.2 in docs/notes about specifying the type of coordinates of the basis function.
- Update section 5 (remove `coord_types` from evals/integrals functions).

The last two changes come from following issue #113 where the keywork `coord_types` became an attribute of the GeneralizedContractionShell class. This implied that instead of this variable being specified in the high level functions (those from the evals and integrals modules) it would be defined when loading a basis set through the function `make_contractions` from the parsers module.


msricher:
What should I do?

gabrielasd:
Hi @msricher and @leila-pujal, I would appreciate your feedback whether the change I did to section 4.2 was OK. Basically keeping the overall structure of section 4 but updating the examples.
I considered whether to merge section 4.2 with 4.1, since the function `make_contraction` was introduced 1st there, but in the end decided to do a smaller change. 

And also, in section 4.2, if I should use a more complex example for the case of `coord_types` being a list or is it OK as it is.

docs/notes.tex
```diff
@@ -400,19 +400,22 @@ \subsection{Types of coordinate systems used by basis functions}
...
-output = high_level_function(basis, coord_type=["cartesian", "spherical"])
+basis = make_contractions(
+    all_basis_dict, ["H", "H"], coords, coord_type=["cartesian", "spherical"]
+)
 \end{lstlisting}
-  In this case, the basis set must consist of exactly two shells of
-  generalized contractions. Otherwise, an error will raised.
+%  In this case, the basis set must consist of exactly two shells of
+%  generalized contractions. Otherwise, an error will be raised.
+In this case, each hydrogen atom must contributes exactly one shell of generalized contractions. Otherwise, an error will be raised.
```
msricher, line 416: Maybe "contributes one list of generalized contractions"? As far as I know, what we call shells are contained in the GeneralizedContractionShell instances.

docs/notes.tex
msricher, line 1: I think having the subsection for coord_type is good, for visibility, since it seems like something people will want to look up.

docs/notes.tex
msricher: I'm not even sure about this, just wondering.

leila-pujal:
Hi all, sorry for the late reply. @gabrielasd thanks for your work. Regarding your questions, I agree with @msricher that we should keep the 4.2 section on its own and not merge it with 4.1. You are right that the make_contractions function is first introduced in 4.1 but the documentation does not go into a lot of explanation of the functions they have been presented. I think it is not redundant to have 4.2, even though it shows again make_contraction, is for a different functionality. I'll keep this open until we merge #113 so we can double-check check we cover all the changes in that PR and if we should explain more about the new API.

gabrielasd:
Got it! Then I'll be pending of PR 113. Thank you @leila-pujal

docs/notes.tex
gabrielasd: I was not sure about this either, but looking at the documentation, the term `shell of generalized contraction"  is defined is section 3.2 as a set of contractions with certain common properties (center, exponents and angular momentum), and then used through the document as such (e.g. section 4.4). So it is a list of generalized contractions like you point out, but with the specific characteristics mentioned. Also in make_contraction the keyword is used to create instances of GeneralizedContractionShell, so I think it is ok to refer to is a such. But let me know if it is more clear in another way.

leila-pujal:
Hi @gabrielasd, could you add also the changes introduced in PR #118? There is a new argument change in the API for some of the high-order functions in density.py, `deriv_type`. You can find more info in the description of the PR but let me know if something is not clear.

gabrielasd:
@leila-pujal I've added the `deriv_type` keyword to the notes.pdf an explained it as a note under the ``Arbitrary derivatives of density" subsection (3.4.2). 
Do you think this is OK or would it be better to dedicate a subsection to explain this, like there was one for types of coordinate systems before?

FarnazH (dismissed):

FarnazH:
@leila-pujal, I hadn't looked at the issue this PR is addressing closely, but It isn't crucial/needed to update this PDF file. The formulas should all be in the docstrings (that @gabrielasd is taking care of), and the code blocks should all be added to the notebooks @maximilianvz and @marco-2023 are working on. Feel free to let me know if I am missing sth.

leila-pujal:
Hi, @gabrielasd it looks perfect what you added and with that now the pdf is updated to the current state of the master branch. @FarnazH, yes you are right that now with the jupyter notbooks it is not crucial to keep this docs updated in the sense that this is not the only place users will come to check Gbasis documentation. But since here there are examples combined with the formulas already made by David I think it looks good that we are updating this file with this PR.

FarnazH (approved):
Thanks @gabrielasd.

Issue #133, opened 2023-10-24
[REQUEST]Gbasis functionality/features jupyter notebooks


leila-pujal:
This issue is to request a set of Jupyter notebooks to show the functionality/features of Gbasis package. This follows recent decisions to incorporate notebooks for qc-devs packages that are easy to use and modify so all the features can be accessed readily.

maximilianvz:
Comment!

FarnazH:
@leila-pujal and @marco-2023 are there any other notebook/example that needs to be added? If so, please list them here.

leila-pujal:
This is issue is completed after merging PRs #172, #158, #147 and #146

PR #142, opened 2023-12-24
`gbasis` Tutorial Notebook

maximilianvz:
## Steps

- [ ] ~Update documentation~
- [ ] Rebase onto master

## Description

In Issue #133, it is being requested that there be a set of Jupyter Notebooks to showcase `gbasis` functionality. In response to this, I have created a tutorial notebook that very closely follows [Toon's notes on the package](https://github.com/theochem/gbasis/blob/5fcd1904475e7cb0b4c4a0d83c1f980de1b42bf7/docs/notes.pdf). Details on the explicit examples that I use are given at the start of the Notebook (a lot of the time, the notes refer to arbitrary systems, but I thought it best to use explicit examples here).

Importantly, I have created the notebook under the impression that PR #140 will be merged - none of the higher-level functions accept a `coord_type` argument.

For some reason, I cannot get some of the equations to render when the Notebook is uploaded to GitHub. However, if you download the Notebook and open it in Jupyter, this can be alleviated when one re-runs the markdown cells containing the problematic equations.

At this stage, the notebook is a bit long but can be divided into three main sections (Basis Sets, Evaluations, Integrals). I can see an opportunity here to divide it into three separate notebooks (one per main section), but this may make things more complicated than they need to be.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |


maximilianvz:
I am going to split this PR into three separate PRs to make the notebooks more digestible.

PR #148, opened 2024-01-10
Update installation, add continuous integration

Ali-Tehrani:
## Steps

- [ ] Update documentation
- [ ] Rebase onto master

## Description
This pull-requests updates how installation is done.  I've removed `setup.py` and replaced it with the modern `pyproject.toml`.  Travis is removed and replaces with github-actions that uses ruff. Added precommit bot to gbasis that runs the various linters.

Also added a coverage comment to the pull-request, where it runs the test and gives out the coverage statement in the form of a comment. For example, see https://github.com/marketplace/actions/pytest-coverage-commentator
Curious to know if this is needed or not.

## Type of Changes

|   | Type |


RichRick1 (approved):
I like these changes a lot!

.editorconfig
```diff
@@ -0,0 +1,18 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
FarnazH, line 1: @Ali-Tehrani, is this file meant to be part of the repo or is it your editor's config file?

.github/workflows/pytest.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  tests:
+    name: "Python ${{ matrix.py }} on OS ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "windows-latest"]
+        py: ["3.7", "3.9", "3.10", "3.11"]
```
FarnazH, line 19: @Ali-Tehrani, is there any reason python 3.8 is not included?

PR #153, opened 2024-01-11
Autograd

RichRick1:
Added automatic differentiation for the contraction. 

1. The function to be used is `eval_nuc_deriv` that has the same API as `_eval_deriv_contractions`.
2. `eval_nuc_deriv` supports arbitrary order derivative with respect to R_x, R_y, R_z (coordinates of the nuclei)
3. Added test to calculate derivatives comparing them with derivatives obtained numerically with `_eval_deriv_contractions` function.

Note: It's vectorized over number of points at which derivative has to be calculated, but function is still looping over primitives and different angular momentum (not components) if more than one is provided.


PaulWAyers:
One can test against @msricher 's `libcint` integration.

marco-2023:
@PaulWAyers This is a stale PR. Do we still want the autograd functionality?

PaulWAyers:
It would be nice to have it, as it lets us do geometry optimization with gbasis, even without libcint. Obviously the libcint should be a preferred implementation, though this may not be so so so so so inefficient.

PR #151, opened 2024-01-10
Fix linters and installation using editable mode

Ali-Tehrani:
This issue fixed the ruff errors. I left it as a pull-request in case it might mess up with merging with other pull-requests. It is also meant to test github actions. I commented out two tests so that the tests pass, see issue #152 for information about the two tests, **these tests have been fixed and now pass.**

## Checklist

- [x ] Write a good description of what the PR does.
- [ ] Rebase onto master

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :hammer: Refactoring  |

PR #156, opened 2024-01-11
Add setup.py to fix install problem

marco-2023:
Gbasis has three folders in its main level and pypi gets lost during the the discovery of top-level packages during the installation. With this the package directory is set.

## Checklist

- [ ] Update documentation
- [x] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,8 @@
+from setuptools import setup
+
+setup(
+    name='your_package',
+    version='0.0',
+    packages=['gbasis'],
+    # other setup parameters...
+)


FarnazH (commented):
@marco-2023, we are using `pyproject.toml` instead of `setup.py`; see https://github.com/theochem/gbasis/pull/148. 
I leave it to @Ali-Tehrani and @RichRick1 to resolve the issue you have encountered.

Ali-Tehrani:
Sorry for causing this to be an issue.  Right now, if you try to install under editable mode it doesn't work: `pip install -e.`  But once this pull-request is merged #151, then you should be able to.  Other than that, I can't imagine what other installation problems can occur, because all of the github actions pass on windows and linux.

marco-2023:
@Ali-Tehrani that is perfect, I will close this PR  and  work on the branch corresponding to PR #151. I noticed it because at the time I could not make a fresh installation.

image

PR #157, opened 2024-01-11
Check the magnitude of negative density values

### LINKED ISSUE #117: [BUG] Negative Electron Density Values

FarnazH:
Following PR #136, the magnitude of negative density values is checked following our discussion in issue #117. 

## Checklist

- [x] Update documentation
- [x] Rebase onto master

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |

## Related
Closes #117


maximilianvz:
This looks good to me, @FarnazH. I think it can be merged.

PaulWAyers (commented):
My only vommrny is that in line 624 of `density.py` the comment should be "check for negative pos. kinetic energy density value" (no "density value" which was appropriate in the first instance but not this one. Otherwise looks great.

FarnazH:
@PaulWAyers, I wanted to keep the comment to one line, that's why I didn't use the whole property name of "negative pos. kinetic energy density", and just sufficed to use density (as it is a density after all).

PR #164, opened 2024-01-24
Add nuclear gradient functions

msricher:
This PR adds nuclear gradients for the overlap integrals, the core Hamiltonian ($\hat{T} + \hat{V}_{ne}$) integrals, and the electron repulsion integrals, in the AO basis.

I based this on pyscf `pyscf/grad/rhf.py`, although that code is for an optimized Hartree-Fock wave function (in the MO basis). Once I am able to test these, the code will be ready to merge. 

## Checklist

- [ ] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |

### DIFF
--- a/gbasis/integrals/libcint.py
+++ b/gbasis/integrals/libcint.py
@@ -1139,7 +1139,7 @@ class CBasis:
         # Compute nuclear gradient
         d_s = self._d_ovlp()
         for iatm, icoords in enumerate(self.atcoords):
-            start, end = self._atm_offs[iatm:iatm + 2]
+            start, end = self._atm_offs[iatm : iatm + 2]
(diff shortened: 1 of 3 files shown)


PaulWAyers:
I would start by adding a test for Cartesian Gaussian primitives. When you take the derivative of a Gaussian primitive with respect to its nuclear coordinate, you pull down a multiplicative factor of $-\alpha | \mathbf{r} - \mathbf{R}_A|$, so each component of the gradient has an extra factor of the corresponding Cartesian dimension. That's an easy test.

The contracted Gaussian functions are similar (just a sum over primitive contributions.

The final bit is the AO -> MO transformation which can be tested vs PYSCF I think.

Issue #101, opened 2019-07-11
[REQUEST] Calculation of density matrices, and spin density matrices


thomaspigeon:
**Is your feature request related to a problem? Please describe.**

thomaspigeon:
I use gbasis to evaluate the spin densities of a molecule. 
The basis informations are first read from a wavefunction file with the help of iodata and the the basis list of tuples is created from the gbasis wrapper with iodata. 
The thing is that to evaluate the spin densities values on a grid, I need the spin density matrices (called one one_dm in the documentation on how to use the function evaluate_density) so I have to recompute the spin density matrices from the alpha/beta coeffs and occupation numbers. I currently do it in a short function.
I am wondering whether it would be useful to add this in gbasis, maybe as an option in the iodata wrapper(or maybe it would be more logical to add it in iodata package). 
What do you think about it ? I could try to make a pull request if you think this is relevant.

FarnazH:
Thanks for your comment and suggestions @thomaspigeon. We have thought about this issue and tried multiple things. The `one_dm` used to be computed by `iodata` in the past, but it was removed because `iodata` is intended to be only a parser module for loading/dumping various file formats (i.e., the `one_dm` might be loaded in one of the formats, but we decided not to include any computations in `iodata` even for sth as simple as `one_dm`)... It also doesn't really belong to `gbasis`, because you need molecular orbitals information to compute density matrices which is not the scope of `gabsis`... So, we believe the `one_dm` computation should be handled in the glue/porcelain code that puts `iodata` and `gbasis` together. This porcelain code would be part of `horton` (which is not available yet) and `chemtools` (which is available, but currently the public `chemtools` still uses `horton2` and it is being changed to use `iodata`, `gbasis`, `grid` & `denspart` modules). In the meanwhile, I think your function for computing `one_dm` is the best solution. Hope this helps.

PaulWAyers:
We already implicitly have the density-matrix and density as "zeroth derivatives" in `gbasis`. So I'm not sure we shouldn't support it.

It's a high-level design decision, really related to how we are going to piece things together. If HORTON is a library of modules with the porcelain that allows them to talk to each other, it is plausible to put this in `gbasis`. It would be nice, probably for numerical efficiency as well as for convenience, if `gbasis` computed the density and its derivatives, and the kinetic energy density and other density-matrix derivatives, and the Fock matrix and such. That's sort of the way it is in the current spec. 

[...]

PR #165, opened 2024-01-29
Add functions for 1rdms and 2-body exchange hole

sanchezw17:
This PR adds functions to evaluate 1-electron rdms and two-body exchange holes as functions of their atomic orbitals. The functions to evaluate the 1rmds are very similar to the preexisting functions which evaluate density; the major difference is the new functions take a list of points as one of their arguments opposed to a single set of points.

I also made small edits in density.py variable names. There were a few occasions where the var_name mentioned in docustrings did not match those used in the actual function bodies.

## Checklist

- [ ] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |


PaulWAyers:
This has showed up in workflows three times in the last year, so I told @sanchezw17 to implement in `gbasis` rather than reinventing the wheel once more. It also addresses #101

msricher (commented):
Looks good to me!

leila-pujal:
hi @sanchezw17, thanks a lot for your PR. I am currently going through the code. I was just testing your new function `evaluate_dm_using_evaluated_orbs`  with an oxygen atom fchk using ccpvtz basis (total of 30 atomic orbitals). At the begning I used a random molecular grid of ~5000 grid points and it tried to allocate 190GB of memory to do the einsum in line 151 of density.py. Probably you already were aware of this because the output array for that function is (Npoints_1, Npoints_2, K_orb, K_orb) but I just wanted to write it here. Maybe we should put a warning or an assertion for a limit on how big the output matrix could be? So einsum does not crash? Please write any thoughts you may have on this topic, or anyone else who is tagged to this PR.

gbasis/evals/density.py
```diff
@@ -625,7 +783,7 @@ def evaluate_posdef_kinetic_energy_density(
     min_output = np.min(output)
```
leila-pujal: I think you forgot to change `output` to `posdef_kindetic_energy_density`. Also, maybe you have a typo there and wanted to use `posdef_kinetic_energy_density`? Just double checking

gbasis/evals/density.py
```diff
@@ -95,12 +95,170 @@ def evaluate_density(one_density_matrix, basis, points, transform=None, threshol
...
+    Returns
+    -------
+    dm_on_grid : np.ndarray(N1,N2,K_orb,K_orb)
+        Density Matrix evaluated at `N1,N2` grid points.
+
+    """
+
+    orb_evals = []
+    # evaluate basi(e)s on the point set(s)
+    for grid in points_list:
```
leila-pujal: You are using a list to pass the points arrays. Lists are mutable objects and I think gbasis tries to reduce the use of them and use more numpy arrays. I was wondering if this this list is ever going to be longer than 2 sets of points? If it is the case it's always going to be either 1 or 2 sets of points (meaning 1 or 2 numpy arrays) I would suggest to provide the sets separately as an argument to the function so you don't have to loop. If not, maybe I would change it to a tuple as it is immutable in python.

gbasis/evals/density.py
```diff
@@ -95,12 +95,170 @@ def evaluate_density(one_density_matrix, basis, points, transform=None, threshol
...
+            raise TypeError(
+                "Evaluation of orbitals must be a two-dimensional `numpy` array with `dtype` float."
+            )
+    if one_density_matrix.shape[0] != one_density_matrix.shape[1]:
+        raise ValueError("One-electron density matrix must be a square matrix.")
+
+    if not np.allclose(one_density_matrix, one_density_matrix.T):
+        raise ValueError("One-electron density matrix must be symmetric.")
+
+    # Tensor product for \gamma(\mathbf{r}_1,\mathbf{r}_2) = \sum_{pq} \gamma_{pq} \chi_p(\mathbf{r}_1) \chi_q(\mathbf{r}_2)
```
leila-pujal: Here probably is that I am not familiar with the math, but based on this equation I would guess the output would be of dimension  `Npoints_1, Npoints_2`, in your notation `k,l`. But you are returning an array ` Npoints_1, Npoints_2, K_orb, K_orb`.  Let me know your thoughts, @sanchezw17, @PaulWAyers

gbasis/evals/density.py
```diff
@@ -95,12 +95,170 @@ def evaluate_density(one_density_matrix, basis, points, transform=None, threshol
...
+    density = np.einsum("ij,ik,jl->klij", one_density_matrix, orb_eval_list[0], orb_eval_list[1])
+
+    # returns dm evaluated on each grid point
+    return density
+
+
+def evaluate_dm_density(one_density_matrix, basis, points_list, transform=None):
+    r"""Return the density of the given basis set at the given points.
```
leila-pujal: If I checked correctly this is the same docstring as in `evaluate_density`. I know they are similar but here two sets of points are used instead of one. Maybe we can add this in the docstring.

gbasis/evals/density.py
```diff
@@ -95,12 +95,170 @@ def evaluate_density(one_density_matrix, basis, points, transform=None, threshol
...
+        dens_evals.append(dens_eval)
+
+    if len(points_list) == 1:
+        dens_evals.append(dens_eval)
+
+    # build density matrix on grid
+    dm_eval = evaluate_dm_density(one_density_matrix, basis, points_list, transform=transform)
+
+    # evaluate hole function
+    numerator = np.einsum("ijkl,ijkl->ijkl", dm_eval, dm_eval)
```
leila-pujal: If I am not wrong, you are multiplying dm_eval * dm_eval right? If this is the case, I was wondering if it would be more efficient to do `dm_eval * dm_eval`? I know using einsum is clear for notation so I am not against it.

leila-pujal (changes requested):
I left comments from my first review. Your new tests pass on my laptop but I haven't checked them in detail. Overall it looks really good to me so if we work out the comments, on my side I will be good to merge it. Thanks again for the PR @sanchezw17.

sanchezw17:
Hi @leila-pujal, thank you for the review. I can add either a warning or a comment at the very least (whichever  you think is most appropriate in this situation). I am still working on implementing these functions, but my idea is to use these functions as callables which are integrated point-by-point, so **i believe** the entire 4D matrix is never stored. At the moment, I do not see the usefulness of storing the entire matrix other than bug-hunting or testing with small bases.

leila-pujal:
Okay I see...just to check if I understood your message you mean you want to use on-the-fly elements of this matrix so you think you are not gonna need to store the whole array? Just to give more information to my message I got the error at this line `density = np.einsum("ij,ik,jl->klij", one_density_matrix, orb_eval_list[0], orb_eval_list[1])` , basically when einsum is trying to allocate/compute the 4D matrix. Thinking that your goal is not to save this matrix and that maybe you just need to generate elements `klij` on the fly, meaning for each pair of orbitals generate `k,l` maybe you could change the code to do subsets of orbitals?. Maybe I am missing some math here but let me know what you think.

sanchezw17:
@leila-pujal you are 100% correct I can have a function pass 2 points instead of 2 sets of points. Since I am in the early stages of working with n_dimentional integration which hasn't been fully tested yet, I want the option to have the full matrix on a grid so it closely matches the shape of "evaluate_density" (which works with grid integration) to help with bug testing. I will add another function that takes two points and returns the density matrix evaluated at that point.

PaulWAyers:
I think we should not try to vectorize over orbitals, but just accumulate the sum over the orbitals. Otherwise the memory is just too much. Or loop over the orbitals whenever the number of grid points is very high.

PaulWAyers:
I think a reasonable API is to read in a set of 3D or 6D points, and then you export the DM/xHOLE on either the 3Dx3D tensor product grid or the DM/xHOLE on the 6D points.

PaulWAyers:
The density matrix is given in terms of its basis set expansion,

$$
\gamma(\mathbf{r},\mathbf{r}') = \sum_{p=1}^{N_{orbs}} \sum_{q=1}^{N_orbs} \gamma_{pq} \phi_p(\mathbf{r})\phi_q(\mathbf{r}')
$$ 

where we have assumed that the orbitals are real. To avoid too much memory, a sensible procedure is:

1. Evaluate $\phi_p(\mathbf{r})$ at the grid points for $\mathbf{r}$.
2. If needed, evaluate $\phi_q(\mathbf{r}')$ at the grid points for $\mathbf{r'}$ (usually, maybe always, the same, so not required).
3. Evaluate density matrix by looping over the orbitals; this saves the memory from forming the gigantic $N_{orbs}^2 \cdot N_{6d-grid}$ object.  (I.e., vectorize over points but not orbitals.) It also allows us to screen by not including a contribution from $\phi_p(\mathbf{r})$ if is centered at a point far from $\mathbf{r}$, but that's a task for the future.

It would be ideal (not required) to allow passing a set of 3-D points (in which case the 6D tensor-product grid is formed internally) or a set of 6-D points (where two 3D sets which may not be the same are appended to each other).

leila-pujal (approved):
Hi @sanchezw17, thanks a lot for updating your code and sorry for the late reply. I reviewed your code and everything looks good. Just a comment, I think it would be more appropriate to use reduced density matrix (rdm) instead of density matrix. For example I would change the name `evaluate_dm_using_evaluated_orbs` to `evaluate_rdm_using_evaluated_orbs` even maybe to be more specific `evaluate_1rdm_using_evaluated_orbs`. Also I think it would be better to specify it on the docstring too so instead of writing density matrix specify first order reduced density matrix. What do you think @FarnazH, @PaulWAyers ?  This is a personal preference suggestion so if you think it is good as it is I don't have any problem. 

I also pushed a commit to change using `numpy.all()` because I was checking the test locally with numpy 2.0 and now np.all for an array of `None` values returns false so your assertion fails (https://numpy.org/doc/stable/reference/generated/numpy.all.html).  

I checked your tests for the first-order reduced density matrix and it looks good but for the exchange hole I don't know why this assertation `np.allclose(eh, np.ones((len(points1), len(points1))) * -1)` . I am sure @PaulWAyers gave you some literature and this is 100% correct but I just wanted to comment that I only checked the test pass. 

I will merge this in the following days probably after I merge the pytest PR #187. Feel free to leave any comments/doubts about my review. Thanks again for working on this.

PaulWAyers:
@sanchezw17 I think @leila-pujal may be correct about `1rdm` instead of `dm`. I often say `dm` but it is nonstandard and less explicit. The number is important; `1dm` would be OK too, but it's important to be clear that it is the one-electron reduced density matrix and so I prefer `1rdm`. I didn't follow the comment on `np.allclose(eh, np.ones((len(points1), len(points1))) * -1)`; probably someone needs to explain that to me in context. I think it is just checking that if the points are identical the hole is -1?  That's only true for same-spin-holes, however.

@sanchezw17 can you "revivify" this so we can merge it in? It's kind of timely right now with IPAM.

PR #171, opened 2024-04-02
Extend screening overlap to IOData wrapper

leila-pujal:
This PR builds on top of PR #111 to extend the screening overlap functionality to data load from and `IOData` object. PR #111 implemented the screening of overlap type integrals. This feature is set up at the user level when the function `make_contractions` is employed in combination with a parsed basis from a Gaussian or NWChem basis set. The same API is added to the function `from_iodata` to be able to access the overlap screening feature. `test_wrappers.py` has been updated and also a new test has been added to check the new API correctly stores the overlap screening information passed when instantiating each `GeneralizedContractionShell`

## Checklist

- [x] Update documentation
- [x] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -142,7 +142,7 @@ def from_iodata(mol, tol=1e-20, overlap=False):
                 shell.kinds[0],
                 icenter=shell.icenter,
                 tol=tol,
-                ovr_screen=overlap
+                ovr_screen=overlap,
             )
         )
 
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -3,13 +3,17 @@ from gbasis.contractions import GeneralizedContractionShell
 import numpy as np
 
 
-def from_iodata(mol):
+def from_iodata(mol, tol=1e-20, overlap=False):
     """Return basis set stored within the `IOData` instance in `iodata`.
 
     Parameters
     ----------
     mol : iodata.iodata.IOData
         `IOData` instance from `iodata` module.
+    tol : float
+        Tolerance used in overlap screening.
+    ovrlap : bool
+        Flag for performing overlap screening between contractions.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


gbasis/wrappers.py
```diff
@@ -3,13 +3,17 @@
...
+def from_iodata(mol, tol=1e-20, overlap=False):
     """Return basis set stored within the `IOData` instance in `iodata`.
 
     Parameters
     ----------
     mol : iodata.iodata.IOData
         `IOData` instance from `iodata` module.
+    tol : float
+        Tolerance used in overlap screening.
+    ovrlap : bool
```
maximilianvz: Typo to fix here ("ovrlap" should be "overlap").

maximilianvz:
@leila-pujal, this looks good to me. The only thing I might suggest is updating the documentation of the `__init__` function of the `GeneralizedContractionShell` class so that the argument names match what is written. Specifically, in [this line](https://github.com/theochem/gbasis/blob/9f6afe7ffc56b9a4e7f8a4a74d0fa5f01197ebca/gbasis/contractions.py#L118), the argument `tol` (given a default value of 1e-15) is referenced in the documentation [here](https://github.com/theochem/gbasis/blob/9f6afe7ffc56b9a4e7f8a4a74d0fa5f01197ebca/gbasis/contractions.py#L144) as `ovr_tol`. This caused a slight bit of confusion for me, because when instantiating an `IOData` shell object as you did [here](https://github.com/theochem/gbasis/blob/373734956bbc9f62bc5ae9e0711302f98a4ca3d7/gbasis/wrappers.py#L144) or when instantiating a `GeneralizedBasisContraction` object as is done in `make_contractions` [here](https://github.com/theochem/gbasis/blob/9f6afe7ffc56b9a4e7f8a4a74d0fa5f01197ebca/gbasis/parsers.py#L251), you obviously need to pass `tol=tol` rather than `ovr_tol=tol` (otherwise, the code will break). When I was trying to cross-reference with the documentation, this difference in naming caught me off guard. You could combine this in a commit with a correction of the typo I pointed out above.

leila-pujal:
Thanks @maximilianvz for your detailed and quick review. I overlooked that typo (probably because in some other places the abbreviation is used). You are right about the mismatch in the documentation. I address it in my last commit as well as the typo (like you suggested). I've asked for another revision from you. Let me know if I forgot to address something from your review.

gbasis/wrappers.py
leila-pujal: Fixed

maximilianvz (approved):
Looks good to me now, @leila-pujal.

PR #179, opened 2024-04-30
Add transform= keyword to cbasis integrals (#178)

### LINKED ISSUE #178: [REQUEST] Make lincint wrapper more flexible
**Is there a change to the code or algorithm you would like to see?**
In the [libcint wrapper](https://github.com/theochem/gbasis/blob/master/gbasis/integrals/libcint.py):
- Add `notation` argument to allow for choosing between `"chemist"` and `"physics"` when computing `electron_repulsion_integral`.
- Add `transform` argument to all integral methods; make sure to keep arguments in the same order as original code. 

These two items can be addressed in separate PRs.

msricher:
Fixes #178: Add transform argument to all integral methods; make sure to keep arguments in the same order as original code.

I don't see where `transform=` keyword is tested in the main Gbasis Python code... @leila-pujal are there tests for this? 

Once I add tests, it can be merged.  I made sure it still runs and passes, at least. I just want to make sure I can replicate any existing tests.

## Checklist

- [x] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/gbasis/integrals/libcint.py
+++ b/gbasis/integrals/libcint.py
@@ -1160,7 +1160,9 @@ class CBasis:
         raise NotImplementedError("Angular momentum integral doesn't work; see Issue #149")
         # return self._amom(origin=origin, notation=notation, transform=transform)
 
-    def point_charge_integral(self, point_coords, point_charges, notation="physicist", transform=None):
+    def point_charge_integral(
+        self, point_coords, point_charges, notation="physicist", transform=None
+    ):
         r"""
         Compute the point charge integrals.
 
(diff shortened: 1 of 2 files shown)


leila-pujal:
@msricher, yes it is not obvious to find where the transformation is applied. So the transform is used in [base_one.py](https://github.com/theochem/gbasis/blob/8df478e7b842cd4b8965034f7039db5623c7e83b/gbasis/base_one.py#L294), [base_two_symm.py](https://github.com/theochem/gbasis/blob/8df478e7b842cd4b8965034f7039db5623c7e83b/gbasis/base_two_symm.py#L405)  and [base_four_symm.py](https://github.com/theochem/gbasis/blob/8df478e7b842cd4b8965034f7039db5623c7e83b/gbasis/base_four_symm.py#L611). Basically it is used when `construct_array_lincomb` method is called for a specific type of base integral (1, 2 or 4 contractions).  I think the functionality is tested for both, higher API functions (for example passing a transfrom to overlap_integral) and for the `construct_array_lincomb()` methods defined in the different base_*.py modules. Let me know if you still have doubts about this and thanks a lot for working on this.

leila-pujal (approved):

PR #182, opened 2024-05-02
Fix incorrect condition for skipping Libcint tests

msricher:
The current test won't work because it only checks if the build files exist: https://github.com/theochem/gbasis/blob/3a88b8400d17070e2e41e43ab281dea5a109279a/tests/test_libcint.py#L62-L63

It should actually check if a file matching `gbasis/integrals/lib/libcint.so*` exists, as this is the actual file that will be loaded, and included with gbasis when it is packaged.

This commit fixes the check, so that the libcint tests will only be skipped if Libcint is missing from the actual package.

@leila-pujal 

## Checklist

- [x] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/tests/test_libcint.py
+++ b/tests/test_libcint.py
@@ -1,11 +1,16 @@
 """Test gbasis.integrals.libcint."""
 
 import pytest
-import os as os
-import sys as sys
+import os
+import sys
+
+from os.path import dirname, join
+from glob import glob
 
 import numpy as np
 import numpy.testing as npt
+
+import gbasis
 
 from gbasis.integrals.angular_momentum import angular_momentum_integral
 from gbasis.integrals.electron_repulsion import electron_repulsion_integral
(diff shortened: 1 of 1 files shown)


leila-pujal:
@msricher, the pytest workflows in github actions don't pass. We get this error:

```
ERROR tests/test_libcint.py - TypeError: glob() got an unexpected keyword argument 'root_dir'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 2.75s ===============================
```

msricher:
You're right, that's not a valid option for older Python versions. This fixes it.

leila-pujal (approved):

PR #185, opened 2024-06-22
Change API for screening overlap integrals

### LINKED ISSUE #169: [QUESTION] Asymmetric overlap correctness
Hello Everyone,
   
I will try to summarize my finding as far as I could.
is_overlap_included: Returns True or False to indicate that these two contractions' overlaps cannot be ignored. It comes in response to whether overlap screening is required by the basis. if required, we verify the possibility by checking whether the shells are far away, and can be screened or not.

If the first contraction overlap screening is not required, we return True. Otherwise, we calculate the distance between the centers of the two shells, and then the cutoff distance (or distance limit) which depends on the shells' minimum exponents and tolerance.
If the distance between the two shells is greater than the cutoff, we return False, otherwise True.

Now, I have two questions in mind regarding overlap and asymmetric overlap in particular. 

1. For Asymmetric overlap /gbasis/integrals/overlap.py Line 185: 
is checking for only the first contraction's ovr_screen sufficient? 
``` bash
if not contractions_one.ovr_screen:
   return True
```
While it does not pose any problems with Symmetric Overlap, but I don't know if the same logic can be applied with Asymmetric Overlap between two contractions from two different basis set. As one of them can be False and the other True?

2. in overlap_integral, I need to ask what does this format in Line 145
``` bash
coord_type = [ct for ct in [shell.coord_type for shell in basis]]
``` 
serve rather than just using 
``` bash
coord_type = [shell.coord_type for shell in basis]
```

FarnazH:
The arguments for screening overlap integrals had to be provided when making a basis (i.e., a list of contractions). Then, when the basis was passed to the `overlap_integrals` function, these arguments were checked to decide whether screening is required. This did not make sense, and meant when adding a screening feature for other integrals, additional arguments had to be added to the `make_contractions` function. This API is changed to request screening when calling the `overlap_integrals` function. This API change naturally fixes the issue raised in issue #169 regarding screening for asymmetric basis.

**Old API:**
```python
# compute overlap integrals without screening
basis = from_iodata(mol, tol=0.001, overlap=True)
olp_ao_screened = overlap_integral(basis)

# compute overlap integrals with screening (default tol is not used)
basis = from_iodata(mol, overlap=False)
olp_ao = overlap_integral(basis)
```

**New API:**
```python
# compute overlap integrals with and without screening (the same basis is used)
basis = from_iodata(mol)
olp_ao = overlap_integral(basis)
olp_ao_screened = overlap_integral(basis, tol_screen=1.0e-6)
```

## Checklist

- [x] Update documentation
- [x] Rebase onto master

## Type of Changes

|   | Type |

## Related
Closes #169


FarnazH:
@leila-pujal, please update the tutorial notebooks accordingly before merging this PR. Thanks.

PaulWAyers (dismissed):
This is *much* better than what we were doing. Thanks @FarnazH . It makes it a lot easier to keep a consistent notation when we use screening for other 1-electron and 2-electron integrals, and even evaluations (like the density and its derivatives).

A GSoC project (that did not get chosen for funding) was going to add screening to these other parts of the code. It's a great project for a new QC-Devs coder or a future GSoC.

leila-pujal (dismissed):
This looks good to me. Thanks for changing the API @FarnazH.

msricher (approved):
This is much cleaner, I think, with the removal of a bunch of duplicated lines in other files, and more consistent for users to have the screening controlled via the Contractions so that all of the evaluations from the same object behave the same.

FarnazH:
@leila-pujal:
1. Could you fix the GitHub Action errors? 
2. Can you please update the notebooks accordinglyand push them here?

leila-pujal:
@FarnazH, sorry I was working on something else before taking care of github actions. I have to open a new PR but it should be straightforward. I think the only problem is for having not supported python version for IOData. I will do that shortly.

leila-pujal (approved):

Issue #121, opened 2021-07-09
[BUG] Big memory usage for density evaluation


leila-pujal:
**Describe the bug**
 
Big memory usage for density evaluation

**To Reproduce**
 
I attach the system I used to test: 
[iron_hexacarbonyl.tar.gz](https://github.com/theochem/gbasis/files/6792948/iron_hexacarbonyl.tar.gz)

```
mol=load_one("iron_hexacarbonyl.fchk")
dm = mol.one_rdms['scf']
basis, coord_types = from_iodata(mol)
becke = BeckeWeights(order=3)

oned = GaussChebyshevType2(100)
rgrid = BeckeTF(1e-4, 1.5).transform_1d_grid(oned)
grid = MolGrid.horton_molgrid(mol.atcoords, mol.atnums, rgrid, 1030, becke)

evaluate_density(dm, basis, grid.points, coord_type=coord_types)
```

**Expected behaviour**

I don't know if this could be improved, I am just reporting the behaviour as an issue. 

**Screenshots**

**System Information:**
 - OS: 18.04.5 LTS
 - Python version: 3.6.10
 - NumPy version: 1.18.1
 - SciPy version: 1.4.1

**Additional context**

[...]

PR #191, opened 2024-06-26
Fix memory usage issue in `evaluate_density` (#121)

msricher:
Fix memory usage issue in `evaluate_density` (#121) by using an optimized `np.einsum`.

## Checklist

- [x] Update documentation
- [x] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### DIFF
--- a/gbasis/evals/density.py
+++ b/gbasis/evals/density.py
@@ -2,6 +2,7 @@
 from gbasis.evals.eval import evaluate_basis
 from gbasis.evals.eval_deriv import evaluate_deriv_basis
 import numpy as np
+import psutil
 from scipy.special import comb
 
 
@@ -57,7 +58,29 @@ def evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval):
             " of the orbital evaluations."
         )
 
-    return np.einsum("pq,qn,pn->n", one_density_matrix, orb_eval, orb_eval, optimize=True)
+    # Output vector of densities
+    density = np.zeros(orb_eval.shape[1], dtype=one_density_matrix.dtype)
+    # Number of chunks (use 8.1 as the number of bytes in a double, just to have a buffer so that we
+    # don't run out of memory). If it fits in 1 chunk, 1 chunk will be used.
+    chunks = max(
+        1,
+        int(
+            8.1
+            * one_density_matrix.shape[0]
+            * orb_eval.shape[1]
+            / psutil.virtual_memory().available
+        ),
+    )
+    # Evaluate densities chunk-wise
+    for density_chunk, orb_eval_chunk in zip(
+        np.split(density, chunks, axis=0), np.split(orb_eval, chunks, axis=1)
+    ):
+        # Evaluate orbital densities
+        d = one_density_matrix.dot(orb_eval_chunk)
+        d *= orb_eval_chunk
+        # Sum orbital densities into output vector
+        np.sum(d, axis=0, out=density_chunk)
+    return density
 
 
 def evaluate_density(one_density_matrix, basis, points, transform=None, threshold=1.0e-8):
--- a/gbasis/evals/density.py
+++ b/gbasis/evals/density.py
@@ -57,9 +57,7 @@ def evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval):
             " of the orbital evaluations."
         )
 
-    density = one_density_matrix.dot(orb_eval)
-    density *= orb_eval
-    return np.sum(density, axis=0)
+    return np.einsum("pq,qn,pn->n", one_density_matrix, orb_eval, orb_eval, optimize=True)
 
 
 def evaluate_density(one_density_matrix, basis, points, transform=None, threshold=1.0e-8):
--- a/gbasis/evals/density.py
+++ b/gbasis/evals/density.py
@@ -6,7 +6,7 @@ import psutil
 from scipy.special import comb
 
 
-def evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval):
+def evaluate_density_using_evaluated_orbs(one_density_matrix, orb_eval, out=None):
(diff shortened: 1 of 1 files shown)


PaulWAyers (commented):
Was it this simple? If so, we should look for other places in the code where the same trick can be used.

How did the performance change? I don't see any reason why this would be slower....

msricher:
> Was it this simple? If so, we should look for other places in the code where the same trick can be used.
> 
> How did the performance change? I don't see any reason why this would be slower....

I think so!  I didn't notice a performance hit. I wouldn't be surprised if it's faster.

msricher:
See https://github.com/theochem/gbasis/issues/121#issuecomment-2192621323. This doesn't make things better, as I found out when I did a real-time profile.

PaulWAyers:
It seemed to good to be true. The memory is that we evaluate *all* the orbitals on a grid first, and then form the density. This is not nearly as good an idea as evaluating directly from the density matrix. **IF** the density matrix is defined in terms of basis functions or localized orbitals, we can screen the evaluation. 

Realistically, this is to change everything. You will not evaluate the basis at first; instead you will, for each grid point, determine whether a given product of basis functions is significant. Once that is known, you will evaluate only the significant products of basis functions. The issue is going to be how to make this efficient given that a "for loop" over grid points is going to be terribly inefficient in Python. I can write some pseudocode for it, but the core issue is going to be making the operation efficient. The good news is that with this done, the cost will be proportional to the number of grid points (independent of the basis) and the memory requirement will be similar to the number of grid points also. 

It is not likely to be able to do this efficiently in the MO basis, so one can really only do this when `transform=None`. (It would work OK for localized MOs, but the logic would be different and more complicated.

PaulWAyers:
With `transform=None`, you can screen. The key idea is to keep track of which atomic-orbitals are needed for a given grid point. 

For a grid point that is $d$ units way from the center of the basis function, the value of the (contracted) basis function is greater than 

$$
c_{\min} n_{\min} d^\ell e^{-\alpha_{min} d^2}
$$

$$
n_{\min} = \sqrt{\frac{\sqrt{2 \alpha_{\min}}(4\alpha_{\min})^\ell}{\sqrt{\pi}(2 \ell + 1)!!}}
$$

where $c_{\min}$, $n_{\min}$, and $\alpha_{\min}$ denote the contraction coefficient, normalization constant, and exponent of the most diffuse Gaussian in a contracted basis function. If the cutoff tolerance is $\epsilon$, you do not need to include this basis function for grid points that are further away than 

$$
\epsilon = c_{\min} n_{\min} d^\ell e^{-\alpha_{min} d^2}
$$

This equation can be solved exactly for $\ell=0$ (in terms of logs) and $\ell=2$ (in terms of the Lambert W function). Otherwise one needs to solve it numerically. For $\ell = 0$,

$$
d =  \left(-\frac{\ln \frac{\epsilon}{ c_{\min} n_{\min}}}{\alpha_{\min}}\right)^{\frac{1}{2}}
$$

For $\ell = 2$ one has the equation,

$$
\frac{- \alpha_{\min} d}{c_{\min} n_{\min}} = -\alpha_{\min}d^2 e^{-\alpha_{min} d^2}
$$

There are two roots, one near the nucleus (where d^2 is small) and one far out. The second root is the one we want, and has the value,

$$
d = \sqrt{\frac{W_{-1}\left(\frac{- \alpha_{\min} d}{c_{\min} n_{\min}}\right)}{- \alpha_{\min}}}
$$

(the same formula, but with the Lambert W function replacing the logarithm. 

We could solve for other values of $\ell$ using a rootfinder, but I don't know that it is worth it. I'd use the logarithm for $\ell = 2$ and the Lambert W otherwise; as long as we aren't too agressive with $\epsilon$ (and we shouldn't be) it will be fine.

The procedure then is to:
1. Group grid points according to the memory available.
2. For each group of grid points, compute a mask array where each shell is either True (included) or False (not included) based on whether it is further than $d$. This mask array has size $n_{grid} \times n_{shells}$. 
3. Evaluate the density but include only contributions from the included shells. This should be possible with a masked-linear-algebra operation. 
4. Continue to other groups of grid points.

In the case where no screening is done, this amounts to a "chunked" version of the old algorithm I think.

ohzde:
I tried to implement this, but I'm not sure if I did it correctly.

I used @msricher [mask](https://github.com/theochem/gbasis/issues/121#issuecomment-2217708347) function, but applied it to each shell individually. I'm using this function to create a boolean array that, for each grid point, indicates whether the evaluation of a specific shell is needed:

```
def _evaluate_orb_mask(shell, points, tol_screen):
    cmin = shell.coeffs.min()
    amin = shell.exps.min()
    nmin = (
            (2 * amin / np.pi) ** (3 / 4)
            * ((4 * amin) ** (shell.angmom / 2))
            / np.sqrt(factorial2(2 * shell.angmom + 1))
        )
    if shell.angmom == 0:
        dist = np.sqrt(-(np.log((tol_screen / (cmin * nmin)) + 0j) / amin).real)
    else:
        dist = np.sqrt(-(lambertw((tol_screen * amin / (cmin * nmin)) + 0j) / amin).real)

    return np.sum((points - shell.coord) ** 2, axis=1) <= dist ** 2 
```

I then passed this mask to `Eval.construct_array_contraction()` and `evaluate_basis()` so that only the relevant shells are evaluated at each point, instead of evaluating all shells at all grid points.

I tested the modified and original versions of the code using the [same example](https://github.com/theochem/gbasis/issues/121#issue-940928029). The modified version runs faster, but the peak memory usage remains the same. These tests were done using the unchunked version of the code.

Old version 
ram_vs_time_rho2

New version:
rho1-new

I feel like I'm doing something wrong because, first, the memory usage increases much faster in the screened version. Also, I'm not sure how accurate the screened densities are.

Issue #192, opened 2024-07-05
[REQUEST] Get pre-commit.ci up and running (without errors)


tovrstra:
**Is your feature request related to a problem? Please describe.**

There is currently a pre-commit configuration, but when running `pre-commit run --all`, one still gets a lot of errors, making it impractical at this stage. It would be good to create a PR in which all these errors are cleaned up, after which we can enable `pre-commit.ci` to run the checks on every PR.

There are probably more useful pre-commit linters to be included. I would recommend including [nb-clean](https://github.com/srstevenson/nb-clean?tab=readme-ov-file#cleaning-pre-commit-hook). It removes all cell outputs and other irrelevant details. These should normally never be included in a Git repository: They take up a lot of space, cause diff noise and can be trivially recomputed.

**Is there a specific integral/formula that you would like implemented?**

No, just CI stuff.

**Is there a change to the code or algorithm you would like to see?**

No, just CI stuff.

**Additional context**

See #187

tovrstra:
@leila-pujal @PaulWAyers @FarnazH 
P.S. I can take care of this if that would be helpful. (Adding nb-clean should be discussed first, and can be done in a later PR.)

leila-pujal:
Thanks @tovrstra for opening the issue. Just to give some background on this, @Ali-Tehrani worked on this in the past in PR #151. When working on the Libcint interface #174 we decided to include the pytest part of #151 but that a new PR should be opened to address all the fixes. That has been on my side but fell through the list of things. I can give it a first try @tovrstra so you don't have to deal with all the changes at first.

tovrstra:
Sounds good. Feel free to ask for a review.

PaulWAyers:
The only issue I see with using `nb-clean` is that we are, in some of the later packages, using notebooks (together with their output) as documentation/tutorials. (Essentially we are replacing what we used to call "scientific documentation" in packages like ChemTools with Jupyter Notebooks.) In those contexts, not including cell outputs is problematic. 

Note that some of the newer packages use the JupyterBook template for building the web site, which has a lot of perks but is probably incompatible with `nb-clean`.

tovrstra:
[...]

PR #137, opened 2023-11-06
Add Libcint bindings

msricher:
## Steps

- [ ✓ ] Write a good description of what the PR does.
- [ ✓ ] Add tests for each unit of code added (e.g. function, class)
- [ ✓ ] Update documentation
- [ ] Squash commits that can be grouped together
- [ ] Rebase onto master

## Description

This is a draft pull request for adding a C interface (to [Libcint](https://github.com/sunqm/libcint)) to GBasis.

Libcint (and most other C(++) codes) work by having a list of atoms, and a list of basis shells, where the atom to which a shell belongs is specified with an index (see [Libcint programmer's manual](https://github.com/sunqm/libcint/blob/master/doc/program_ref.pdf)). In GBasis (and IOData), I noticed that this concept is not followed, and is combined into one List[GeneralizedContractionShell] structure. This is fine, but there are some problems:

### 1
The `Attributes` section of the GeneralizedContractionShell documentation indicates that it should have a `charge` attribute, but this seems to be unused (it doesn't exist). 

### 2
The IOData `Shell` class has this attribute:
```
 icenter
        An integer index specifying the row in the atcoords array of IOData object.
```
@leila-pujal and I have worked on exposing `icenter` in the GeneralizedContractionShell object, but I might need more help with that since I'm not sure where it's supposed to have been assigned, when running the test I wrote.

(Note: this has been fixed. There was confusion about whether certain quantities (atomic {numbers,coordinates}) were accessible through GBasis objects or passed as arguments to the integral functions.)
### Remaining tasks

* (DONE) ~~Figure out how to give to my bindings the `icenter` and `charge` attributes of each atom/shell it needs.~~
* Test and fix the bindings.
* Include some more robust testing for all C-binded integrals.
* Finalize the interface by which the C bindings are accessed instead of the Python functions.
* Squash all the commits, rebase PR.

Can anyone give me some advice on this? (Also it's my first time making a PR, so I hope I followed the guidelines well enough.)

## Type of Changes

|   | Type |
| ------------- | ------------- |
|   | :bug: Bug fix  |
| ✓ | :sparkles: New feature |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

PR #193, opened 2024-07-10
Fix pre-commit errors

leila-pujal:
This PR addresses the request posted on issue #192 and includes the missing parts of PR #151 that were not included in #137. This also relates to PR #187. 

## Checklist

- [ ] ~~Squash commits that can be grouped together~~
- [x] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# EditorConfig is awesome: https://EditorConfig.org
+# EditorConfig: https://EditorConfig.org
 
 root = true
 
--- a/gbasis/base_four_symm.py
+++ b/gbasis/base_four_symm.py
@@ -2,9 +2,10 @@
 import abc
 import itertools as it
 
+import numpy as np
+
 from gbasis.base import BaseGaussianRelatedArray
 from gbasis.spherical import generate_transformation
-import numpy as np
 
 
 # pylint: disable=W0235
--- a/gbasis/base_one.py
+++ b/gbasis/base_one.py
@@ -1,9 +1,10 @@
 """Base class for arrays that depend on one contracted Gaussian."""
 import abc
 
+import numpy as np
+
 from gbasis.base import BaseGaussianRelatedArray
 from gbasis.spherical import generate_transformation
-import numpy as np
 
 
 # pylint: disable=W0235
--- a/gbasis/base_two_asymm.py
+++ b/gbasis/base_two_asymm.py
@@ -1,9 +1,10 @@
 """Base class for arrays that depend on two contracted Gaussians."""
 import abc
 
+import numpy as np
+
 from gbasis.base import BaseGaussianRelatedArray
 from gbasis.spherical import generate_transformation
-import numpy as np
 
 
 # pylint: disable=W0235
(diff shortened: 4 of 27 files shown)


leila-pujal:
Hi @tovrstra, I made the PR to include all the errors from running `pre-commit run --all`. I was able to fix most of them but there are some @msricher and me are unsure about how impactful they are. I list them below:
```
gbasis/integrals/libcint.py:185:16: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
gbasis/integrals/libcint.py:194:16: RUF012 Mutable class attributes should be annotated with `typing.ClassVar`
gbasis/integrals/libcint.py:314:17: B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
gbasis/integrals/libcint.py:811:21: RUF005 Consider iterable unpacking instead of concatenation
gbasis/integrals/libcint.py:1236:13: B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling

```
@msricher commented that we don't use any annotation in Gbasis (I think IOData and Grid does) and then the one we are more unsure about it is the `B904` with the error exception. Any thoughts on how we should handle these would be super useful. Thanks!

tovrstra:
Ruff has recommendations for every error message it can generate and these are usually quite informative: https://docs.astral.sh/ruff/rules/

I'll try to give some advice for each type of problem;

`RUF012`: annotations are generally a useful thing, but I can understand you don't want to get into it now. Adding them everywhere for consistency is a lot of work and outside this scope of this PR. You may try making the `_fields_` attribute a tuple instead of a list, e.g. as in the following example from the ctypes documentation:

```python
class POINT(Structure):
    _fields_ = ("x", c_int), ("y", c_int)
````

I'm not sure if that will solve the Ruff message, but it is a useful change anyway. There is no reason to have a mutable `_fields_` attribute. Feel free to add an ignore line to `pyproject.toml` for `RUF012` if it keeps complaining.

`B904`: In both cases, it would suggest deviating from the EAFP pattern, i.e. something as follows instead:

```python
cfunc = self._cache.get(attr)
if cfunc is None:
    cfunc = self._get_cfunc(attr)
    self._cache[attr] = cfunc
return cfunc
```

Inside the helper method `_get_cfunc`, you can raise the exception if needed. As a side effect, the linter message will go away. I don't follow the argument that EAFP always leads to more readable code. In simple examples, it sometimes does. By working with the `get` method, you also address the EAFP argument that it reduces the number of dictionary lookups.

`RUF005`: Can be fixed as `out_shape = (self.nbfn, self.nbfn, self.nbfn, self.nbfn, *components)`. See https://docs.astral.sh/ruff/rules/collection-literal-concatenation/

tovrstra:
P.S. If you prefer to stick to EAPF, which is certainly fine, then you can fix `B904` as follows:

```python
raise ValueError(f"there is no ``gbasis`` API for the function {attr}") from None`
```

The more elaborate option would be:

```python
try:
     cfunc = self._cache[attr]
except KeyError as exc:
    ...
    else:
       raise ValueError(f"there is no ``gbasis`` API for the function {attr}") from exc
```

The second form seems less suited here because the KeyError is related to the caching and will not offer useful information.

PR #196, opened 2024-08-02
Fix compatibility with changes in IOData (prepare_segmented)

### LINKED ISSUE #195: [BUG] `from_iodata` not working after API change of `iodata`
The API change introduced in https://github.com/theochem/iodata/pull/373 breaks the [`from_iodata`](https://github.com/theochem/gbasis/blob/3e063ad3a3ce40430ac1eafeffc6f92e79888ff1/gbasis/wrappers.py#L89) function. So, `gbasis` does not work with the latest source code of `iodata`.

tovrstra:
The required function is imported locally from the iodata library because the wrapper module also contains a wrapper for pyscf. When the import is made at the top and IOData is not installed, it would also become impossible to use the pyscf wrapper. (You may want to split the wrapper module into one per package to avoid local imports, but this is not critical.)

This commit includes a few side effects:

- Ruff fixes were required in wrappers.py to make the commit.
- .envrc is added to .gitignore. I used direnv for managing environments

## Checklist

- [x] Tests pass locally
- [x] Pre-commit passes
- [x] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related

Fixes #195

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ celerybeat-schedule
 *.sage.py
 
 # Environments
+.envrc
 .env
 .venv
 env/
--- a/gbasis/wrappers.py
+++ b/gbasis/wrappers.py
@@ -1,8 +1,8 @@
 """Module for interfacing to other quantum chemistry packages."""
 
-from gbasis.contractions import GeneralizedContractionShell
 import numpy as np
 
+from gbasis.contractions import GeneralizedContractionShell
 
 CONVENTIONS_LIBCINT = {
     (5, "p"): ["s5", "s4", "s3", "s2", "s1", "c0", "c1", "c2", "c3", "c4", "c5"],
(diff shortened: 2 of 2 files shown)


leila-pujal (approved):
Hi @tovrstra, thanks a lot for your quick PR to fix issue #195. I checked your new code locally with the latest IOData with one of my scripts and it works. All tests that use `from_iodata` pass as well. You may be right and we should split pyscf and iodata wrappers now that we have this import specific to IOData. I think maybe because  `from_pyscf` is not a big function they decided to put both into one module and of course, they were not importing any specific functionality. I will open an issue as an enhancement so we can keep track of this suggestion. Related to your side effects: The Ruff changes match the ones I did in `wrappers.py` in #193 . I will merge this PR and update #193 so no problem. Thanks for adding the .envrc to .gitignore . I am not sure how new `direnv` is but we should keep track better of new tools for managing enviromental variables and add them to .gitingnore. I will soon deal with PR #187 and #193. I will merge this at the end of the day to give more time to Canadian people if they want to review it as well. Thanks again!

tovrstra:
You're welcome. Thanks for the review. I'll let you do the merging, so you can decide in which order to deal with the PRs.

Issue #207, opened 2025-02-14
GSoc 2025: Improve Performance Using Screening


PaulWAyers:
## Description
Use screening based on the overlap between basis functions to improve performance.

## :books: Package Description and Impact
For large molecules, textbook expressions for quantities expanded in Gaussian basis functions (e.g., the electron density) or integrals based on Gaussian basis functions (e.g., the kinetic-energy integral) typically include many negligible terms. By screening out these terms, and only evaluating terms that are nonnegligible, the performance of `GBasis` can be greatly enhanced. 

## :construction_worker: What will you do?
 In `GBasis` we provide a utility for screening these terms using their overlap, [`is_overlap_included`](https://github.com/theochem/gbasis/blob/9f6afe7ffc56b9a4e7f8a4a74d0fa5f01197ebca/gbasis/integrals/overlap.py#L156). When this expression is small, one can also neglect other one-electron integrals. A generalization of this approach allows fast evaluation of spatial quantities and 2-electron integrals. Your main goal would be to screen 1-electron integrals and the evaluation of quantities at (grid) points using overlap screening and its generalization. 

## :checkered_flag: Expected Outcomes
1. Adapt `is_overlap_included` to screen other one-electron integrals. 
2. Extend `is_overlap_included` to three functions, which allows screening spatial evaluations.
3. Write tests to ensure correctness and assess performance.
4. :trophy: An ambitious stretch goal is to implement screening of 2-electron integrals.

| | |
| --- | ----------- |
| Required skills | Python, OOP |
| Preferred skills | Be comfortable with math, physics. Experience with scientific programming,  quantum chemistry would be huge plus |
| Project size | 350 hours, Large |
| Difficulty | Medium  😉 |

## :raising_hand: Mentors
| | | |
| --- | ----------- | -- |
| Marco Mart&iacute;nez-Gonz&aacute;lez | mmg870630_at_gmail_dot_com  | @marco-2023 |
| Esteban V&ouml;hringer-Martinez | estebanvohringer_at_qcmmlab_dot_com   | @evohringer  |
| Paul Ayers| ayers_at_mcmaster_dot_ca  | @PaulWAyers  |
| Gabriela Sánchez-Díaz| sanchezg_at_mcmaster_dot_ca  | @gabrielasd   |

## 📝 Notes & References
- Supersedes #167 
- Related to #183 , #121, 
[...]

PR #210, opened 2025-06-11
Two-index screening

ohzde:
## Description

This PR adds screening feature to Gbasis for cases involving two contractions. It adds screening of overlaps, kinetic energy, arbitrary-order moments, momentum, and angular momentum integrals via the `tol_screen` parameter. A new module, `screening`, has been created, with `two_index_screening()` as its first function.

## Checklist

- [ ] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |

## Related
Issues #207

### DIFF
--- a/gbasis/integrals/angular_momentum.py
+++ b/gbasis/integrals/angular_momentum.py
@@ -1,11 +1,13 @@
 """Module for evaluating the integral over the angular momentum operator."""
+import numpy as np
+
 from gbasis.base_two_symm import BaseTwoIndexSymmetric
 from gbasis.contractions import GeneralizedContractionShell
 from gbasis.integrals._diff_operator_int import (
     _compute_differential_operator_integrals_intermediate,
 )
 from gbasis.integrals._moment_int import _compute_multipole_moment_integrals_intermediate
-import numpy as np
+from gbasis.screening import two_index_screening
 
 
 # TODO: need to test against reference
@@ -56,17 +58,22 @@ class AngularMomentumIntegral(BaseTwoIndexSymmetric):
     """
 
     @staticmethod
-    def construct_array_contraction(contractions_one, contractions_two):
+    def construct_array_contraction(contractions_one, contractions_two, tol_screen=None):
         """Return the integrals over the angular momentum operator of the given contractions.
 
         Parameters
         ----------
         contractions_one : GeneralizedContractionShell
             Contracted Cartesian Gaussians (of the same shell) associated with the first index of
-            the kinetic energy integral.
+            the angular momentum integral.
         contractions_two : GeneralizedContractionShell
             Contracted Cartesian Gaussians (of the same shell) associated with the second index of
-            the kinetic energy integral.
+            the angular momentum integral.
+        tol_screen : None or float, optional
+            The tolerance used for screening angular momentum integrals. The `tol_screen` is combined with the
+            minimum contraction exponents to compute a cutoff which is compared to the distance between
+            the contraction centers to decide whether the angular momentum integral should be set to zero (i.e.,
+            screened). If `None`, no screening is performed.
 
         Returns
         -------
@@ -98,6 +105,18 @@ class AngularMomentumIntegral(BaseTwoIndexSymmetric):
             raise TypeError("`contractions_one` must be a `GeneralizedContractionShell` instance.")
         if not isinstance(contractions_two, GeneralizedContractionShell):
             raise TypeError("`contractions_two` must be a `GeneralizedContractionShell` instance.")
+
+        if two_index_screening(contractions_one, contractions_two, tol_screen):
+            return np.zeros(
+                (
+                    contractions_one.num_seg_cont,
+                    len(contractions_one.norm_prim_cart),
+                    contractions_two.num_seg_cont,
+                    len(contractions_two.norm_prim_cart),
+                    3,
+                ),
+                dtype=complex,
+            )
 
         diff_integrals = _compute_differential_operator_integrals_intermediate(
             1,
@@ -156,7 +175,7 @@ class AngularMomentumIntegral(BaseTwoIndexSymmetric):
         return -1j * np.transpose(output, (3, 2, 4, 1, 0))
 
 
-def angular_momentum_integral(basis, transform=None):
+def angular_momentum_integral(basis, transform=None, tol_screen=None):
     r"""Return the integral over :math:`hat{L}` of the given basis set.
 
     .. math::
@@ -183,6 +202,11 @@ def angular_momentum_integral(basis, transform=None):
         Transformation is applied to the left, i.e. the sum is over the index 1 of `transform`
         and index 0 of the array for contractions.
         Default is no transformation.
+    tol_screen : None or float, optional
+        The tolerance used for screening angular momentum integrals. The `tol_screen` is combined with the
+        minimum contraction exponents to compute a cutoff which is compared to the distance between
+        the contraction centers to decide whether the angular momentum integral should be set to zero (i.e.,
+        screened). If `None`, no screening is performed.
 
     Returns
     -------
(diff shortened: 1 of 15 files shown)


gbasis/screening.py
```diff
@@ -0,0 +1,52 @@
+"""1 and 2-index screening functions"""
+
+import numpy as np
+
+
+def two_index_screening(contractions_one, contractions_two, tol_screen):
```
marco-2023: A more expressive name is preferable, such as `is_two_index_integral_screened`, to better convey the purpose of the function. Function names should denote an action. Also, this is a low-level function; it is better for `tol_screen` to only accept a float. Handling None to disable screening is better done by the functions that use this.

pyproject.toml
```diff
@@ -100,7 +100,7 @@ line-length = 100
 # PL is pylint
 # RUF is Ruff-specific rules
 select = ["E", "F", "UP", "B", "I", "PGH", "PL", "RUF"]
-line-length = 100
+line-length = 120
```
marco-2023: `line-length = 100` is the format adopted by the organization. Before a PR the code should be linted using `black -l 100`. Please don't change this.

gbasis/integrals/angular_momentum.py
```diff
@@ -99,6 +106,18 @@ def construct_array_contraction(contractions_one, contractions_two):
         if not isinstance(contractions_two, GeneralizedContractionShell):
             raise TypeError("`contractions_two` must be a `GeneralizedContractionShell` instance.")
 
+        if two_index_screening(contractions_one, contractions_two, tol_screen):
```
marco-2023: Add logic outside for the no-screening case (`None`)

gbasis/integrals/overlap.py
```diff
@@ -161,58 +164,11 @@ def overlap_integral(basis, transform=None, tol_screen=None):
     """
     coord_type = [shell.coord_type for shell in basis]
     kwargs = {"tol_screen": tol_screen}
+
```
marco-2023, line 174: Each integral should have a default (usually appropriate) cutoff value `tol_screen` different. Because of this I think kwargs should be two arguments i.e. `screen_basis : bool` and `tol_screen:float` then we can fix the default values for both for each type of integrals

marco-2023 (changes requested):

gbasis/integrals/angular_momentum.py
ohzde: added. please let me if there is any issue with this

gbasis/integrals/overlap.py
ohzde: I added `screen_basis` and set the default `tol_screen` to 1e-8 for all integrals for now. please let me know if I'm handling the unscreened case, screening with the default tolerance, or screening with an arbitrary tolerance incorrectly

pyproject.toml
ohzde: there were a few lines (formulas for kinetic energy, moment, and angular momentum) that already exceeded the 100 limit, which is why I changed the limit to 120. this time, I tried breaking those formulas into multiple lines. if there's any issue with this, please let me know.

marco-2023:
Hi @ohzde, I've added some fixes to the PR. I added them myself because these were not major changes.

1. Changed the name of the screening function. We will apply more than one screening, so we need names to differentiate them.
2. Fixed tests: Some overlap tests were failing because the api of the function was not updated (None was used for no screening).
3. Added a Screening precision test for overlap.
4. Fixed tests: For all two-index related tests, screening was used, as this is the default now. 

#### TODO:

- Please take a look at the changes, we will need to do this for all tests as we add basis screening for other methods
- Please check the precision test added for overlap and add the equivalent test for kinetic energy, momentum, ... etc. This is the only test targeting the effects of screening.

tests/test_angular_momentum.py
```diff
@@ -411,3 +411,22 @@ def test_angular_momentum_integral_lincomb():
         angular_momentum_integral_obj.construct_array_lincomb(transform, ["spherical"]),
         angular_momentum_integral(basis, transform),
     )
+
+
+@pytest.mark.parametrize("precision", [1.0e-5, 1.0e-6, 1.0e-7, 1.0e-8])
+def test_angular_momentum_screening_accuracy(precision):
+    """Test angular momentum screening.
+
+    This test is meant to  fail.  Using cartesian sto-6G nwchem basis set.
```
marco-2023: why this comment?

tests/test_kinetic_energy.py
```diff
@@ -249,3 +249,21 @@ def test_kinetic_energy_integral_horton_anorcc_bec():
     )
     ke_int_values = kinetic_energy_integral(basis, screen_basis=False)
     assert np.allclose(ke_int_values, horton_kinetic_energy_integral)
+
+@pytest.mark.parametrize("precision", [1.0e-5, 1.0e-6, 1.0e-7, 1.0e-8])
+def test_kinetic_screening_accuracy(precision):
+    """Test kinetic energy screening.
+
+    This test is meant to  fail.  Using cartesian sto-6G nwchem basis set.
```
marco-2023: Why this comment?

tests/test_moment.py
```diff
@@ -322,3 +322,23 @@ def test_moment_spherical_lincomb():
             screen_basis=False,
         ),
     )
+
+
+@pytest.mark.parametrize("precision", [1.0e-5, 1.0e-6, 1.0e-7, 1.0e-8])
+def test_moment_screening_accuracy(precision):
+    """Test (dipole) moment screening.
+
+    This test is meant to  fail.  Using cartesian sto-6G nwchem basis set.
```
marco-2023: why this comment?

tests/test_momentum.py
```diff
@@ -206,3 +206,22 @@ def test_momentum_integral_lincomb():
         momentum_integral_obj.construct_array_lincomb(transform, ["spherical"], screen_basis=False),
         momentum_integral(basis, transform=transform, screen_basis=False),
     )
+
+
+@pytest.mark.parametrize("precision", [1.0e-5, 1.0e-6, 1.0e-7, 1.0e-8])
+def test_momentum_screening_accuracy(precision):
+    """Test momentum screening.
+
+    This test is meant to  fail.  Using cartesian sto-6G nwchem basis set.
```
marco-2023: comment?

tests/test_overlap_asymm.py
```diff
@@ -105,3 +106,22 @@ def test_overlap_integral_asymmetric_compare():
             screen_basis=False,
         ),
     )
+
+
+@pytest.mark.parametrize("precision", [1.0e-5, 1.0e-6, 1.0e-7, 1.0e-8])
+def test_overlap_asymmetric_screening_accuracy(precision):
+    """Test asymmetric overlap screening.
+
+    This test is meant to  fail.  Using cartesian sto-6G nwchem basis set.
```
marco-2023: comment?

tests/test_overlap.py
```diff
@@ -229,8 +229,8 @@ def test_overlap_screening_vs_without_screening():
     overlaps_no_screen = overlap_integral(contraction, tol_screen=None)
     assert np.allclose(overlaps, overlaps_no_screen)
 
-
-def test_overlap_screening_with_fail():
+@pytest.mark.parametrize("precision", [1.0e-5, 1.0e-6, 1.0e-7, 1.0e-8])
+def test_overlap_screening_tight(precision):
     """Test overlap screening.
 
     This test is meant to  fail.  Using cartesian sto-6G nwchem basis set.
```
marco-2023: Please remove the comment here, too. My bad

marco-2023 (changes requested):

tests/test_overlap.py
ohzde: fixed, sorry about this!

marco-2023 (dismissed):

marco-2023:
This looks good to me now.

PaulWAyers (dismissed):
LGTM too

marco-2023:
I had to add some missing fixes to the libcint tests. Now it should be ok.

marco-2023 (approved):

PaulWAyers (approved):
Thanks for being careful @marco-2023

PR #213, opened 2025-09-27
Test direct density derivatives

marco-2023:
The tests for the electron density derivative functions were adapted such that the cases  `"direct`"  and general `"general"` are tested using the same tests. The module with the `"direct`" density derivative tests was removed as it is no longer needed (it was duplicated code).

## Checklist

- [ x] Write a good description of what the PR does.
- [ x] Update documentation
- [ x] Squash commits that can be grouped together
- [ x] Rebase onto master

## Type of Changes

|   | Type |


PaulWAyers (approved):
LGTM

PR #226, opened 2026-01-21
Force the use of old jupyter book as temporal fix

marco-2023:
GitHub Actions defaults to Jupyter Book v2, which is incompatible with the current website configuration. This change forces the workflow to use Jupyter Book v1.

## Checklist

- [x] Update documentation
- [ ] Rebase onto master

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |


PaulWAyers (approved):
LGTM

PR #234, opened 2026-02-27
[PR-2] feat: implement VRR  for OS+HGP two-electron integral pipeline

San1357:
## Summary
- Implemented `_vertical_recursion_relation()` — VRR (Eq. 65), first recursion step of OS+HGP pipeline
- Builds angular momentum on center A: `[a+1,0|00]^m` from `[00|00]^m` using Boys function values
- NumPy vectorized operations with precomputed coefficients for efficiency
- Part of Issue #221 (Week 2)

## What Changed

| File | Lines | What |
|------|-------|------|
| `gbasis/integrals/_two_elec_int_improved.py` | +130 | VRR implementation |
| `tests/test_two_elec_int_improved.py` | +102 | 4 VRR tests |

## Mathematical Reference

The Vertical Recursion Relation (Eq. 65) is:

$$[a+1_i, 0 | 00]^m = (P-A)_i \, [a0|00]^m - \frac{\rho}{\zeta}(Q-P)_i \, [a0|00]^{m+1} + \frac{a_i}{2\zeta}\left([a-1_i, 0|00]^m - \frac{\rho}{\zeta}[a-1_i, 0|00]^{m+1}\right)$$

where $\zeta = \alpha_a + \alpha_b$, $\rho = \zeta\eta/(\zeta+\eta)$, $P$ is the Gaussian product center, and $m$ is the auxiliary index from the Boys function.

## How To Test

```bash
# Run VRR tests
pytest tests/test_two_elec_int_improved.py -v


# Check coverage
pytest tests/test_two_elec_int_improved.py --cov=gbasis.integrals._two_elec_int_improved --cov-report=term-missing


# Check formatting & linting
black --check --line-length 100 gbasis/integrals/_two_elec_int_improved.py
ruff check gbasis/integrals/_two_elec_int_improved.py

# Verify existing tests not broken
pytest tests/ --timeout=120

```

### Tests (15 passed)
image

## Check coverage
image

### Black: Clean
image

### Ruff: Clean
image

### Existing tests: Not broken
295 passed, 279 skipped, 1 xfailed, 4 warnings in 67.30s 

## Checklist
- [x] Tests added for each new function
- [x] All 4 tests pass
- [x] Coverage 100%
- [x] Black formatted (line-length 100)
- [x] Ruff linting clean
- [x] Existing tests not broken (295 passed)
- [x] Docstrings in numpydoc format
- [x] No new dependencies added (uses existing numpy)
- [x] Pure Python only (no C/Fortran)

## Scope of this PR
[...]

Issue #221, opened 2025-12-01
WoC:  2-electron integrals: better performance and more flexibility


PaulWAyers:
## Description

Add functionality to the `GBasis` library for evaluating 2-electron integrals, including non-Coulomb repulsions. The current Python implementation is very slow, but it also uses a *very* inefficient algorithm. A more efficient algorithm, combining the Obara-Saika and Head-Gordon-Pople recursions, would fix this issue. One advantage of having this coded in (pure) Python is that it makes it easy to explore new screening methods, exploit sparsity, and include unconventional integrals (e.g., screened Coulomb interactions). Here is a good introduction to [Gaussian integral evaluation](https://joshuagoings.com/2017/04/28/integrals/), which can help you get your bearings without overwhelming you with (important!) details.

## :books: Package Description and Impact

`GBasis` is a pure-Python package for evaluating and analytically integrating Gaussian-type orbitals and their related quantities. The goal is to build a set of tools to the quantum chemistry community that are easily accessible and easy to use as to facilitate future scientific works. 

## :construction_worker: What will you do?

Your main focus will be to (re)implement the 2-electron integrals in `GBasis`. This is important partly because it seems we may have a bug in the integrals for high angular momentum (cf. #216) and because the current implementation is painfully slow. It is very helpful for many people to have a pure Python implementation of the 2-electron integrals because, while it will always be (relatively) slow, it allows more flexibility for development of new algorithms and facilitates cross-platform code interoperability. In addition, in many cases integral evaluation is not the performance bottleneck.

The key steps will be to
1. Understand the recursive evaluation of Gaussian integrals.
2. Implement the key recursions. (vertical, horizontal, electron-transfer)
3. Apply these recursions in sequence to evaluate the 2-electron integrals.
4. Write functions for different "initializations" of the Boys function to support different types of 2-electron integrals.

2-electron integrals from [PySCF](https://pyscf.org/) can be used for testing, as can a recursion check against existing integrals (which are surely reliable for low angular momentum). For more general integral types, one will have to test by manual verification (e.g., numerical integration using [Grid](https://github.com/theochem/grid).

## :checkered_flag: Expected Outcomes

1. Implement the recursive evaluation of 2-electron integrals into `GBasis` using the algorithm in the notes. This requires implementing several different recursions, which are the "building blocks" for the algorithm as a whole. 
2. Implement several different "generalized Boys functions" so that multiple types of 2-electron integrals are supported.
[...]

PR #237, opened 2026-03-04
[PR-3] feat: implement ETR (Eq. 66) and primitive contraction for OS+HGP pipeline

San1357:
## Summary
- Implemented `_electron_transfer_recursion()` — ETR (Eq. 66), transfers angular momentum from bra to ket side
- Implemented `_optimized_contraction()` — primitive-to-contracted contraction using `np.einsum`
- Implemented `_get_factorial2_norm()` — normalization helper using double factorial
- Part of Issue #221 (Week 3),
-  builds on PR #234 (VRR)

## What Changed

| File | Lines | What |
|------|-------|------|
| `gbasis/integrals/_two_elec_int_improved.py` | +150 | ETR + contraction implementation |
| `tests/test_two_elec_int_improved.py` | +80 | 5 ETR + norm tests |

## Mathematical Reference

The Electron Transfer Recursion (Eq. 66) is:

$$[c+1_i, d | a, b] = [c, d | a+1_i, b] + (A-C)_i [c, d | a, b]$$

where angular momentum is transferred from center A to center C.

## How To Test

```bash
# Run tests
pytest tests/test_two_elec_int_improved.py -v

# Check coverage
pytest tests/test_two_elec_int_improved.py --cov=gbasis.integrals._two_elec_int_improved --cov-report=term-missing

# Check formatting & linting
black --check gbasis/integrals/_two_elec_int_improved.py
ruff check gbasis/integrals/_two_elec_int_improved.py

# Verify existing tests not broken
pytest tests/ --timeout=120
```

### Tests (9 passed)

image

## Check coverage
image

### Black: Clean & Ruff: Clean
image

### Existing tests: Not broken
300 passed, 279 skipped, 1 xfailed, 4 warnings in 67.86s (0:01:07)

## First Checklist
- [x] Tests added for each new function
- [x] All 9 tests pass
- [x] Black formatted
- [x] Ruff linting clean
- [x] Existing tests not broken
- [x] Docstrings in numpydoc format
- [x] No new dependencies added (uses existing numpy)
- [x] Pure Python only (no C/Fortran)

## Scope of this PR
- This PR adds ETR and contraction as part of the OS+HGP pipeline (Week 3)
- Adds `_electron_transfer_recursion` (ETR)
- Adds `_optimized_contraction` with per-center vectorized norm computation
- Adds `functools.cache` to `_get_factorial2_norm` for performance
- Does NOT modify any existing code
- Does NOT change any existing tests
- Next step: HRR (Week 4)

## Second Checklist
- [ ] Update documentation
- [x] Rebase onto master

## Type of Changes

|   | Type |
|   | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### DIFF
--- a/gbasis/integrals/_two_elec_int_improved.py
+++ b/gbasis/integrals/_two_elec_int_improved.py
@@ -1,0 +1,386 @@
+"""Improved two-electron integrals using Obara-Saika + Head-Gordon-Pople recursions.
+
+This module implements VRR, ETR, and primitive contraction steps of the
+OS+HGP algorithm for two-electron integrals.
+
+Algorithm overview (full pipeline):
+1. Start with Boys function F_m(T) for m = 0 to angmom_total
+2. VRR: Build [a0|00]^m from [00|00]^m (Eq. 65)        <-- Done (PR 2)
+3. ETR: Build [a0|c0]^0 from [a0|00]^m (Eq. 66)        <-- THIS PR
+4. Contract primitives                                    <-- THIS PR
+5. HRR: Build [ab|cd] from [a0|c0] (Eq. 67)             <-- Future
+
+References:
+- Obara, S. & Saika, A. J. Chem. Phys. 1986, 84, 3963.
+- Head-Gordon, M. & Pople, J. A. J. Chem. Phys. 1988, 89, 5777.
+- Ahlrichs, R. Phys. Chem. Chem. Phys. 2006, 8, 3072.
+"""
+
+import numpy as np
+
+from gbasis.utils import factorial2
+
+# Cache for factorial2 values to avoid repeated computation
+_FACTORIAL2_CACHE = {}
+
+
+def _get_factorial2_norm(angmom_components):
+    """Get cached factorial2 normalization for angular momentum components.
+
+    Parameters
+    ----------
+    angmom_components : np.ndarray(n, 3)
+        Angular momentum components.
+
+    Returns
+    -------
+    norm : np.ndarray(n,)
+        Normalization factors 1/sqrt(prod((2*l-1)!!)).
+    """
+    key = tuple(map(tuple, angmom_components))
+    if key not in _FACTORIAL2_CACHE:
+        _FACTORIAL2_CACHE[key] = 1.0 / np.sqrt(
+            np.prod(factorial2(2 * angmom_components - 1), axis=1)
+        )
+    return _FACTORIAL2_CACHE[key]
+
+
+def _optimized_contraction(
+    integrals_etransf,
+    exps_a,
+    exps_b,
+    exps_c,
+    exps_d,
+    coeffs_a,
+    coeffs_b,
+    coeffs_c,
+    coeffs_d,
+    angmom_a,
+    angmom_b,
+    angmom_c,
+    angmom_d,
+):
+    """Optimized primitive contraction using einsum.
+
+    Parameters
+    ----------
+    integrals_etransf : np.ndarray
+        ETR output with shape (c_x, c_y, c_z, a_x, a_y, a_z, K_d, K_b, K_c, K_a).
+    exps_a/b/c/d : np.ndarray
+        Primitive exponents.
+    coeffs_a/b/c/d : np.ndarray
+        Contraction coefficients.
+    angmom_a/b/c/d : int
+        Angular momenta.
+
+    Returns
+    -------
+    contracted : np.ndarray
+        Contracted integrals with shape (c_x, c_y, c_z, a_x, a_y, a_z, M_a, M_c, M_b, M_d).
+    """
+    # Precompute normalization constants (1D arrays)
+    norm_a = (2 * exps_a / np.pi) ** 0.75 * (4 * exps_a) ** (angmom_a / 2)
+    norm_b = (2 * exps_b / np.pi) ** 0.75 * (4 * exps_b) ** (angmom_b / 2)
+    norm_c = (2 * exps_c / np.pi) ** 0.75 * (4 * exps_c) ** (angmom_c / 2)
+    norm_d = (2 * exps_d / np.pi) ** 0.75 * (4 * exps_d) ** (angmom_d / 2)
+
+    # Multiply coefficients by normalization (more efficient than per-element)
+    coeffs_a_norm = coeffs_a * norm_a[:, np.newaxis]
+    coeffs_b_norm = coeffs_b * norm_b[:, np.newaxis]
+    coeffs_c_norm = coeffs_c * norm_c[:, np.newaxis]
+    coeffs_d_norm = coeffs_d * norm_d[:, np.newaxis]
+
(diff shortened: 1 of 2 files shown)


gbasis/integrals/_two_elec_int_improved.py
```diff
@@ -0,0 +1,386 @@
...
+- Head-Gordon, M. & Pople, J. A. J. Chem. Phys. 1988, 89, 5777.
+- Ahlrichs, R. Phys. Chem. Chem. Phys. 2006, 8, 3072.
+"""
+
+import numpy as np
+
+from gbasis.utils import factorial2
+
+# Cache for factorial2 values to avoid repeated computation
+_FACTORIAL2_CACHE = {}
```
msricher: Python has a simple tool for this type of caching since Python 3.9(??):
https://docs.python.org/3.14/library/functools.html#functools.cache

You can just decorate your function:
```python3
@functools.cache
def _get_factorial2_norm(angmom_components):
   ...
```

gbasis/integrals/_two_elec_int_improved.py
```diff
@@ -0,0 +1,386 @@
...
+    """
+    key = tuple(map(tuple, angmom_components))
+    if key not in _FACTORIAL2_CACHE:
+        _FACTORIAL2_CACHE[key] = 1.0 / np.sqrt(
+            np.prod(factorial2(2 * angmom_components - 1), axis=1)
+        )
+    return _FACTORIAL2_CACHE[key]
+
+
+def _optimized_contraction(
```
msricher: I wonder if a better function signature would just be `def f(integrals, exps, coeffs, angmoms):`?

Then you can use it like this, if all of the exps, coeffs, and angmoms come from indexing the corresponding arrays:
`f(integrals, exps[(a, b, c, d),], coeffs[(a, b, c, d),], angmoms[(a, b, c, d),])`

I'm not sure if the angmoms are extracted from an array, though... but in general using fewer function arguments by applying NumPy's advanced indexing, or by other grouping strategies, is cleaner. Is it faster? Not sure, but the difference is probably not much.

Just some thoughts I had reading this part.

gbasis/integrals/_two_elec_int_improved.py
```diff
@@ -0,0 +1,386 @@
...
+    angmom_a/b/c/d : int
+        Angular momenta.
+
+    Returns
+    -------
+    contracted : np.ndarray
+        Contracted integrals with shape (c_x, c_y, c_z, a_x, a_y, a_z, M_a, M_c, M_b, M_d).
+    """
+    # Precompute normalization constants (1D arrays)
+    norm_a = (2 * exps_a / np.pi) ** 0.75 * (4 * exps_a) ** (angmom_a / 2)
```
msricher: This part would definitely benefit from using arrays and vectorized operations with NumPy.

gbasis/integrals/_two_elec_int_improved.py
```diff
@@ -0,0 +1,386 @@
...
+            )
+            if m_max > 2:
+                integrals_etransf[:, :, c + 1, :, :, 1:-1, ...] = (
+                    qc_plus_zoe_pa_z * integrals_etransf[:, :, c, :, :, 1:-1, ...]
+                    + a_coeff_z * integrals_etransf[:, :, c, :, :, :-2, ...]
+                    + c_coeff * integrals_etransf[:, :, c - 1, :, :, 1:-1, ...]
+                    - zeta_over_eta * integrals_etransf[:, :, c, :, :, 2:, ...]
+                )
+
+    return integrals_etransf
```
msricher, line 366: The rest of it, the actual contractions and recurrences, seems good to me.

gbasis/integrals/_two_elec_int_improved.py
San1357: Hi @msricher  
Done! Replaced `_FACTORIAL2_CACHE = {}` with `@functools.cache` decorator.
Since `functools.cache` requires hashable arguments and NumPy arrays are not hashable, the function now accepts a tuple of tuples instead of an array.

gbasis/integrals/_two_elec_int_improved.py
San1357: Done! Simplified the signature as suggested:

`_optimized_contraction(integrals_etransf, exps, coeffs, angmoms)`

where:
- `exps = (exps_a, exps_b, exps_c, exps_d)`
- `coeffs = (coeffs_a, coeffs_b, coeffs_c, coeffs_d)`  
- `angmoms = (angmom_a, angmom_b, angmom_c, angmom_d)`

The values are unpacked inside the function.
is this ok? @msricher

gbasis/integrals/_two_elec_int_improved.py
San1357: @msricher 
### For suggestion #3 (vectorizing line 82) -
The ETR loops have sequential dependencies where each step `c+1` uses the previous step `c`, so full vectorization isn't straightforward. 

Could you clarify which specific part you had in mind?

gbasis/integrals/_two_elec_int_improved.py
San1357: Thank you for the feedback! @msricher

gbasis/integrals/_two_elec_int_improved.py
msricher:
```python3
angmoms = np.array([l_a, l_b, l_c, l_d])
exps = np.array([e_a, e_b, e_c, e_d])
coeffs = np.array([c_a, c_b, c_c, c_d])
# etc. ...
norms = ((2 / np.pi) * exps) ** 0.75 * (4 * exps) ** (angmoms / 2)
coeffs_norm = coeffs * norms[:, :, np.newaxis] # or something like this, there's some
                                               # broadcasting you can do to make it work
```

I just mean that if you are going to pass these arguments in as single arrays, then there's a bit more vectorization you can do with the (a, b, c, d) centers.

San1357: Understood! @msricher,
I was confused, now I've Vectorized the norm computation over all 4 centers as suggested. Now using NumPy arrays and computing norms in one shot using broadcasting — all 4 centers (a, b, c, d) are handled together instead of separately.

The caller-side passing of single arrays (e.g. exps[(a,b,c,d),]) will be handled in the full pipeline when HRR is integrated. For now, the conversion to arrays is done internally inside the function. 

Does this look good to you? @msricher

San1357: Actually after testing , I realized that using np.array stacking breaks when shells have different numbers of primitives (K). For example, if K_a=2 and K_b=1, np.array([exps_a, exps_b]) creates a ragged array and fails. So I've used per-center list comprehension instead — the norm computation is still vectorized using NumPy ops per center, just not across all 4 centers simultaneously

marco-2023:
Superseeded by #245

PR #251, opened 2026-06-01
add basis sets

Aydin-Bagheri:
Hello,
I have uploaded atomic basis sets (from QuasiAos repository), which are needed for population analysis with QUAO/IAO methods.


PaulWAyers (commented):
I think ano-rcc is already there with a different name and ano-rcc.gbs.bak is redundant.

Aydin-Bagheri:
> I think ano-rcc is already there with a different name and ano-rcc.gbs.bak is redundant.

I have updated the files.

PR #258, opened 2026-08-09
PR-[10] - feat: add Windows CI support 

San1357:
## Summary

Add Windows CI support to the pytest workflow.

## Changes

- Added `windows-latest` to the OS matrix in `.github/workflows/pytest.yaml`
- Added separate install steps for Windows vs non-Windows runners
- Windows uses `pip install --no-build-isolation` with `cmake` and `ninja` pre-installed

## Testing

CI will verify that the build and tests pass on Windows with Python 3.9, 3.10, 3.11, and 3.12.

## Related

- Part of PR-10 (Windows CI support)
- Complements existing Ubuntu and macOS CI coverage


msricher (dismissed):
Maybe check if a newer SciPy version requirement ("scipy>=1.18.0"?) would fix the issue on Windows... NumPy 2.0 was a major change there and they've fixed stuff since then.

tests/test_libcint.py
msricher, line 615: Can you make sure the arguments, keywords, etc., match, between the gbasis Python and CBasis implementation? For example, `coord_type` and `coord_types` do not match here.

San1357: Thanks for catching this, The difference is intentional — make_contractions takes coord_types (plural) as per the standard GBasis API, while CBasis.__init__ takes coord_type (singular) to match the libcint convention. Both are correct for their respective functions.

